PR #57, opened 2019-03-27
Include original traceback when unwrapping an error

thedrow:
This will allow the users to understand better where the exception came from and why it occurred.

### DIFF
--- a/returns/result.py
+++ b/returns/result.py
@@ -59,7 +59,10 @@ class Failure(Result[Any, _ErrorType]):
 
     def unwrap(self):
         """Raises an exception, since it does not have a value inside."""
-        raise UnwrapFailedError(self) from self._inner_value
+        if isinstance(self._inner_value, BaseException):
+            raise UnwrapFailedError(self) from self._inner_value
+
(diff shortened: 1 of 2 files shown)


returns/result.py
```diff
@@ -59,6 +59,9 @@ def value_or(self, default_value):
 
     def unwrap(self):
         """Raises an exception, since it does not have a value inside."""
+        if isinstance(self._inner_value, BaseException):
```
sobolevn: We do not work with `BaseException`, only `Exception`.

sobolevn:
And a test, please 🙂

thedrow:
Not sure how to test this correctly...

sobolevn:
Hey, @thedrow! You can use `with pytest.raises(UnwrapFailedError) as excinfo` and then get its traceback: https://docs.python.org/3/library/traceback.html

thedrow:
I'll happify the linter in a second.

sobolevn:
Awesome work, thanks! 🎉 

P.S. Are you interested in #60? 😄

PR #86, opened 2019-06-13
WIP: rewrites types for Result monad

### LINKED ISSUE #85: Many types are incorrect
After a review done by @astynax it was revealed that many types are incorrect.

We have privately discussed all the problems and solutions.
Here's the short version:
1. We should only deal with `Result` itself, `Success` and `Failure` should return `Result` instances
2. We should not work with `Success` and `Failure` directly, since there's no such thing, only `Result`
3. `Result` functions should always return `Result` changing only the type signature

Btw, @astynax is now a core-dev of `returns` library. Congrats :tada:

sobolevn:
Please, check my direction.
Closes #85


sobolevn:
Only docs are remaining.

sobolevn:
@astynax please, review everything. Now it should be 100% complete.

Issue #47, opened 2019-02-27
Use `pytest-mypy-plugins` to write typed tests


sobolevn:
https://github.com/mkurnikov/pytest-mypy-plugins

sobolevn:
Closed with https://github.com/dry-python/returns/pull/88

PR #88, opened 2019-06-14
Mypy plugin for typed decorators

sobolevn:
```python
from returns.result import safe

@safe
def test() -> int:
    return 1

reveal_type(test)
reveal_type(test())
```

When running `mypy` it yields correct results:

```
» mypy test.py --show-traceback
test.py:7: error: Revealed type is 'def () -> returns.result.Result[builtins.int*, builtins.Exception]'
test.py:8: error: Revealed type is 'returns.result.Result[builtins.int*, builtins.Exception]'
```

I have not tested it yet, I will need to implement #47 first, but this is huge!
Since I am really annoyed by this issue. I had so many mistakes already in my project because of this limitation.


sobolevn:
@proofit404 
> I suggest to move it into returns.contrib.mypy.decorator_plugin.

Good idea!

> I strongly suggest using https://github.com/mkurnikov/pytest-mypy-plugins before the merge.

Sure! That's what I have mentioned in the PR's body:
> I have not tested it yet, I will need to implement #47 first, but this is huge!

PR #217, opened 2019-12-19
flake8 exclude updated

pvldruzhinin:
Extended virtualenvs list in setup.cfg

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,9 +32,14 @@ exclude =
   # Trash and cache:
   .git
   __pycache__
-  .venv
   .eggs
   *.egg
+  # Environments
+  .env
+  .venv
+  env/
+  venv/
+  ENV/
 
 ignore = D100, D104, D401, W504, X100, WPS202, WPS214, WPS320, WPS440, TAE002, RST303, RST304, DAR103, DAR203
 


setup.cfg
```diff
@@ -32,9 +32,14 @@ exclude =
   # Trash and cache:
   .git
   __pycache__
-  .venv
   .eggs
   *.egg
+  # Environments
```
sobolevn: After researching this topic a bit, I am going to reject this change. Here's why:
- `flake8` does not feature this on purpose: https://gitlab.com/pycqa/flake8/issues/579 and https://gitlab.com/pycqa/flake8/issues/36 There are two many possible ways to go wrong. There are also `.venv-SOME_HASH`, `.venv3.7`, `.venv3`, `.venv3.7.5`, etc

- It can be easily fixed on the users side. You can rename `venv` into `.venv` and done! `poetry` by default uses `.venv` when configured with `virtualenvs.in-project = true`

So, thanks a lot for this suggestion and these knew information I got!

pvldruzhinin:
Thanks, this is great answer to another question "what's the 'best' venv folder name". I'll update rREADME then.

sobolevn:
@pvldruzhinin sure! Thanks!

PR #218, opened 2019-12-19
Updated CONTRIBUTING.md: flake8's default .venv folder excluding

pvldruzhinin:


sobolevn:
@pvldruzhinin Thanks! 

You can also update `CONTRIBUTING.md` with the same line:
- Here: https://github.com/wemake-services/wemake-python-package/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/CHANGELOG.md
- And here: https://github.com/wemake-services/wemake-python-package/blob/master/CONTRIBUTING.md

PR #226, opened 2020-01-20
RequiresContext container implementation

### LINKED ISSUE #216: Consider adding Reader monad
Useful pieces of information:
- https://gist.github.com/dustingetz/3172127
- https://dev.to/gcanti/getting-started-with-fp-ts-reader-1ie5
- https://github.com/dbrattli/OSlash/blob/master/tests/test_reader.py
- https://github.com/dbrattli/OSlash/wiki/Three-Useful-Monads#the-reader-monad

It might be used for type-safe DI and context-based computations.
It might be a good addition.

### LINKED ISSUE #230: Use doctests where possible

### LINKED ISSUE #231: Should we make IO lazy?
```ts
interface IO<A> {
  (): A
}

const now: IO<number> = () => new Date().getTime()

/** write to the `console` a random boolean */
const program: IO<void> = io.chain(randomBool, log)

program()
```

> Note that nothing happens until you call program().

> That's because program is a value which just represents an effectful computation, so in order to execute any side effect you must "run the IO action".

I have firstly made `IO` direct, so it executes as soon as it is called. Should we revisit this statement?

sobolevn:
I would be thrilled to get your feedback, guys!

P.S. For some reason I cannot request a review from @pvldruzhinin and @theLastOfCats, but I do request it!

Closes #225 
Closes #216 
Closes #230
Closes #231


theLastOfCats:
LGTM

pvldruzhinin:
All looks good!

sobolevn:
TODO for later:

- [ ] support `join`
- [ ] add readme example
- [ ] add docs to webpage
- [ ] add API spec to webpage
- [ ] add typetests for `context`

sobolevn:
I would also love to experiment with `Reader` that can hold both functions and values.

pvldruzhinin:
> Didn't work. Python type system does not allow this.

Maybe you need smth like this?

def curry(aFunction):
	""" 
	Turns a normal python function into a curried function.

	Most easily used as a decorator when defining functions:
		@curry
		def add(x, y): return x + y

	"""
	funcName = aFunction.__code__.co_name
	numArgs = aFunction.__code__.co_argcount
	def buildReader(argValues, numArgs):
		if (numArgs == 0): return aFunction(*argValues)
		else:
			return lambda x: buildReader(argValues + [x], numArgs - 1)
	return Reader(buildReader([], aFunction.__code__.co_argcount))

sobolevn:
@pvldruzhinin sorry, I didn't get your advice. How should this help me with type inference of python objects and callables?

pvldruzhinin:
> @pvldruzhinin sorry, I didn't get your advice. How should this help me with type inference of python objects and callables?

Yeah, I have no full understanding of the problem. Just tried to find difference between our code and PyMonad implementation.
But try to look at all Reader implementation in PyMonad (edited link in previous answer), it seems they support functions and values.

sobolevn:
Because they are untyped. Let me show where the problem is.

Current implementation:

```python
class RequiresContext(
    BaseContainer,
    Generic[_EnvType, _ReturnType],
):

    def __init__(
        self, inner_value: Callable[[_EnvType], _ReturnType],
    ) -> None: ...
```

With this annotation `mypy` is able to infer correct types of both environment and return type.

When we change the constructor method type to `def __init__(self, inner_value: _ReturnType)`, then we won't have any `_EnvType` to work with. And it will be set to `<nothing>`

And this is not what we want.

pvldruzhinin:
Ah, i see, thanks for the explanation 👌

PR #299, opened 2020-03-27
Fix typo in `Maybe` docstring

homeworkprod:
I *think* this should say "computations" rather than "commutation".

Should it?


sobolevn:
Thanks a lot!

Issue #278, opened 2020-03-16
Consider `Maybe` methods that state if a value is present/absent


homeworkprod:
Hi,

I'm just getting my feet wet with `returns`. As a first iteration, I'm trying to introduce `Maybe` into a somewhat larger codebase of mine.

One thing that irks me a bit is code like this:

Coming from code like this:
```py
item: Optional[Item] = find_item(item_id)
if item is None:
    raise SomeCustomException("ohai")
```

(Naively) porting to `Maybe` leaves me with this:
```py
item: Maybe[Item] = find_item(item_id).value_or(None)
if item is None:
    raise SomeCustomException("ohai")
```

While I can replace `is not None` checks after which I want to process the value with `Maybe.map` (and not raise an exception if the value is missing) and gain more succinct code, almost the opposite seems to be the case if I don't want to work with the potential value but rather its existence itself.

Another example:
```py
# traditional
token: Optional[Token] = find_token(some_key)
valid_token_found: bool = (token is not None) and not token.is_expired

# `returns`-y way
from returns.maybe import Some
token: Maybe[Token] = find_token(some_key)
valid_token_found: bool = token.map(lambda t: not t.is_expired) is Some
```

**What I'm looking for** is a method like `[is_]{present, absent, some, none}`, `has_value`, `empty`, etc. to allow me to do this:
```py
# `returns`-y way
# NB: No import necessary!
token: Maybe[Token] = find_token(some_key)
valid_token_found: bool = token.map(lambda t: not t.is_expired).present
```

Am I overlooking some best practice that avoids imports and identity/equality operators? If not, are you open to consider adding such (potentially even more auto-complete-friendly) methods?

Thanks!

sobolevn:
Hi @homeworkprod! Thanks a lot for your interest in `returns`!

Considering your feature request, by design Python does not allow us to override `is` operation. Because, it should serve the only purpose: tell if object is the same one in memory. We cannot override this.

But, I see another ways to achieve your goals here:

```python
from returns.maybe import Maybe
from returns.pipeline import is_successful

token: Maybe[Token] = find_token(some_key)
valid_token_found = is_successful(token.map(lambda t: not t.is_expired))
```

And:

```python
item: Maybe[Item] = find_item(item_id)
if not is_successful(item):
    raise SomeCustomException("ohai")
```

Does this work for you?

homeworkprod:
Hi @sobolevn, thank you for your quick response.

My intent is not to override `is` for this use case.

[...]

Issue #95, opened 2019-06-19
Override __bool__ function for Maybe and Result


sobolevn:

sobolevn:
This can cause confusion: some containers will be convertable to `bool` while other is not.
Better to leave `__bool__` as is: always `True`.

thedrow:
I'm reopening this since it's convenient to do the following:
```python
if maybe:
  value = maybe.unwrap()
else:
  pass # handle this case
```

sobolevn:
Also related:
- #301 
- #300

homeworkprod:
This seems like an interesting approach.

It depends a lot on the name of the evaluated variable/callable to make sense to the reader, though, I think.

thedrow:
Naming variables is one of the first things you learn when you program (hopefully)

sobolevn:
Ok, I've made a decision: we should leave `is_successful` function as the main and only to tell whether container was successful or not.

Reasoning:

1. `__bool__` is too implicit. It is hard to tell: is `bool(Success(False))` a `True` or a `False`? Supporting `__bool__` also opens a door of overusing `if container:` conditions.
2. Custom method like `is_success` and `is_failure` requires either duplicating `is_successful` or removing it. I don't see a reason for duplication or a breaking change here.
3. `is_successful` on the other hand limits what can be passed in there on a type level. It is already there and costs only a single import. 

So, it is going to stay as is. Thanks a lot for participating in the discussion and proposing lots of good alternatives!

PR #300, opened 2020-03-27
Add `Maybe.{is_some, is_nothing}`

homeworkprod:
As suggested and discussed in #278.

Contains docstring-based tests. Is this sufficient?

### DIFF
--- a/returns/maybe.py
+++ b/returns/maybe.py
@@ -162,6 +162,32 @@ class Maybe(
         """
         raise NotImplementedError
 
+    def is_some(self) -> bool:
+        """
+        Returns ``True`` only if this ``Maybe`` is a ``Some``.
+
+        .. code:: python
+
+          >>> from returns.maybe import Nothing, Some
+          >>> assert Some(1).is_some()
+          >>> assert not Nothing.is_some()
+
+        """
+        raise NotImplementedError
+
+    def is_nothing(self) -> bool:
+        """
+        Returns ``True`` only if this ``Maybe`` is a ``Nothing``.
+
+        .. code:: python
+
+          >>> from returns.maybe import Nothing, Some
+          >>> assert not Some(1).is_nothing()
+          >>> assert Nothing.is_nothing()
+
+        """
+        raise NotImplementedError
+
     @classmethod
     def lift(
         cls,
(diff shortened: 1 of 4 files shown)


homeworkprod:
Alright, checks are finally green!

> The only question I have: what about `Result`?
> Can we create an uniform interface for it too?

I have opened #301 for that; please take a look.

homeworkprod:
@sobolevn, how can we continue from here?

And is #301 what you had in mind, or something more "uniform" as in the same method names in both `Maybe` and `Result`?

homeworkprod:
@sobolevn, is there anything I can do to move this forward? Are you unsure about something?

I'd like to have something like this before continuing to integrate the library into my project.
And the upcoming long Easter weekend is a good time for me to work on that :)

sobolevn:
@homeworkprod yes, I am a bit unsure about new methods in our classes. The thing is I don't really like methods. I like functions. 

And we already have one to solve this problem. And this PR introduces logic duplication in exchange of better Developer Experience. That's always hard to reason about 🙂

homeworkprod:
Isn't the whole point of this library a better developer experience?

thedrow:
In this case, I'd prefer to have a single flag and not two.

Will `can_unwrap` fit?

homeworkprod:
> In this case, I'd prefer to have a single flag and not two.

I do see the convenience of that.

However, as I've laid out in [my comment](https://github.com/dry-python/returns/issues/278#issuecomment-602203163), separate methods/functions is what "Maybe" implementations in popular programming languages' standard libraries provide.

There are quite a few things *returns* does in its own ways, but I think it's worth to consider what people wanting to (or having to) use it can expect given their prior experiences.

homeworkprod:
@sobolevn, I'm surprised that you're hesitating with this suggestion as you have marked my comment with a heart (which then made me go on and submit this PR). I guess you just changed your mind afterwards?

sobolevn:
@homeworkprod you idea is awesome. We are just thinking about the best naming / design for this exact feature. There are different options:

1. `is_successful` as it is know, because fewer methods is better
2. Unified getters for all types: `Result`, `Maybe`, `IOResult`, `TaskResult`, etc
3. Different methods for different containers as suggested here and in #301 
4. `__bool__` cast

There's also no need to hurry. New release is still in heavy development and we will ship this feature in it.

I really appreciate you involvement in this project! Thanks a lot for raising issues, discussing ideas, and sendings PRs! 👍

homeworkprod:
Thanks for clarifying!

sobolevn:
Thanks a lot for the PR! I am going to close it, since we plan to support another API.

See https://github.com/dry-python/returns/issues/95#issuecomment-629625442

There's also an additional way to do the same thing: `isinstance(value, Maybe.success_type)`

PR #301, opened 2020-03-29
Add `Result.{is_success, is_failure}`

homeworkprod:
Idea suggested by @sobolevn in https://github.com/dry-python/returns/pull/300#pullrequestreview-382804352.

### DIFF
--- a/returns/result.py
+++ b/returns/result.py
@@ -243,6 +243,32 @@ class Result(
         """
         raise NotImplementedError
 
+    def is_success(self) -> bool:
+        """
+        Returns ``True`` only if this ``Result`` is a ``Success``.
+
+        .. code:: python
+
+          >>> from returns.result import Failure, Success
+          >>> assert Success(1).is_success()
+          >>> assert not Failure(1).is_success()
+
+        """
+        raise NotImplementedError
+
+    def is_failure(self) -> bool:
+        """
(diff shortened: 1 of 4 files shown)


thedrow:
I prefer to use `successful` and `failed` which is proper English.

If you prefer you can use `is_successful` and `has_failed`.

sobolevn:
All these method can be replaced with `__bool__` which you proposed in #95

homeworkprod:
> I prefer to use successful and failed which is proper English.

You seem to suggest that my recommendations are not proper English. I would politely disagree. "is *a* success" would be preferable, but since this is an API and not a text, and I think it's okay to use this (common) shortcut.

The naming pattern is aligned with what I suggested in #300 (i.e. `is_some` and `is_nothing`), which itself is based on what other programming languages do.

Of course it's fine to consider other names, but that feels like bikeshedding to me. After all, I value consistency, which in turn allows to derive names easily without having to look at the docs.

I'm sorry if this might read as aggressive; it is not meant to be.

thedrow:
This is why I like Ruby's convention for boolean methods: `success?`.

sobolevn:
Thanks a lot for the PR! I am going to close it, since we plan to support another API.

See https://github.com/dry-python/returns/issues/95#issuecomment-629625442

There's also an additional way to do the same thing: `isinstance(value, Result.success_type)`

PR #337, opened 2020-04-20
Issue overload

### LINKED ISSUE #330: curry plugin does not work with `Overloaded` type

sobolevn:
Closes #330


sobolevn:
I want to preserve history, so not squashing it.

PR #348, opened 2020-04-22
Don't merge me: Currying

orsinium:


orsinium:
Sorry, that's not supposed to be here, yet.

PR #354, opened 2020-04-25
Add `NoDeps` type alias for `Any`

### LINKED ISSUE #273: Add `NoDeps` type alias for `Any`
Sometimes `RequiresContext` and other similar types might be used with no explicit dependencies (due to some existing API requirements for example).

So, we need to have `NoDeps` type alias for `Any`.

thepabloaguilar:
The type alias was created in `returns/context/requires_context.py` and reexported in `returns/context/__init__.py` file.

The type of "empty" attribute in `RequiresContext` was changed, from:
```python
class RequiresContext:
    # code
   empty: ClassVar[Any] = object()
```
to:
```python
class RequiresContext:
    # code
    empty: ClassVar[NoDeps] = object()
```

The same modifications was made in `RequiresContextResult` and `RequiresContextIOResult` classes.

Closes #273

### DIFF
--- a/returns/context/requires_context.py
+++ b/returns/context/requires_context.py
@@ -27,6 +27,9 @@ _ErrorType = TypeVar('_ErrorType')
 # Helpers:
 _FirstType = TypeVar('_FirstType')
 
+# Type Aliases:
+NoDeps = Any
+
 
 @final
 class RequiresContext(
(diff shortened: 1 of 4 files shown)


returns/context/requires_context.py
```diff
@@ -27,6 +27,9 @@
 # Helpers:
 _FirstType = TypeVar('_FirstType')
 
+# Type Aliases:
```
sobolevn, line 30: Can you please document it with `#:` 
This way it will be visible to sphinx. 

We also need to describe why do we have this type alias.

thepabloaguilar: Done!

PR #357, opened 2020-04-27
Creates "not_" function

### LINKED ISSUE #313: Add `not_` function
It should work like so:

```python
assert not_(is_successful)(Success(1)) is False
assert not_(is_successful)(Failure(1)) is True
```

Related #300 
Related #301

thepabloaguilar:
I've implemented the function and it's working properly. I've wrote doc tests and examples in documentation, as well.

Closes #313

### DIFF
--- a/returns/functions.py
+++ b/returns/functions.py
@@ -1,3 +1,4 @@
+from functools import wraps
 from typing import Any, Callable, NoReturn, TypeVar
 
 # Aliases:
@@ -144,3 +145,11 @@ def raise_exception(exception: Exception) -> NoReturn:
 
     """
     raise exception
+
+
+def not_(function: Callable[..., bool]) -> Callable[..., bool]:
+    """Denies the function returns."""
+    @wraps(function)
+    def wrapped_function(*args, **kwargs) -> bool:  # noqa: WPS430
+        return not function(*args, **kwargs)
+    return wrapped_function
--- a/returns/functions.py
+++ b/returns/functions.py
@@ -159,8 +159,8 @@ def not_(function: Callable[..., bool]) -> Callable[..., bool]:
       ...     return isinstance(result_container, Result.success_type)
       ...
 
-      >>> assert not_(is_successful)(Success(1)) is False
-      >>> assert not_(is_successful)(Failure(1.0)) is True
+      >>> assert not_(is_successful)(Success(1.0)) is False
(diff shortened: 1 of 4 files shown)


thepabloaguilar:
@sobolevn Can you take a look, please?

returns/functions.py
```diff
@@ -144,3 +148,12 @@ def raise_exception(exception: Exception) -> NoReturn:
 
     """
     raise exception
+
+
+def not_(
```
sobolevn, line 153: I would say that `not_` has different type.

It is more like: `def not_(function: Callable[..., bool]) -> Callable[..., bool]:`
This function can take any args and kwargs from the callee function.

We can tweak it in `mypy` plugin to work correctly with type (currently it is required).

thepabloaguilar: The implementation will be like this:
```python
def not_(not_(function: Callable[..., bool]) -> Callable[..., bool]:
    return lambda *args, **kwargs: not function(*args, **kwargs)
```
right?

returns/functions.py
```diff
@@ -144,3 +148,12 @@ def raise_exception(exception: Exception) -> NoReturn:
...
+    function: Callable[
+        [Result[_ValueType, _ErrorType]], Union[_Containers, bool],
+    ],
+) -> Callable[[Result[_ValueType, _ErrorType]], bool]:
+    """Denies the function return."""
+    return lambda container: not bool(function(container))
```
astynax, line 159: `functools.wraps`?

thepabloaguilar: Maybe it is a great ideia, and to be more readable we have to re-write. It will look like this:
```python
def not_(function: Callable[..., bool]) -> Callable[..., bool]:
    """Denies the function return."""
    @wraps(function)
    def wrapped_function(*args, **kwargs):
        return not function(*args, **kwargs)
    return wrapped_function
```

thepabloaguilar: @sobolevn I think this is a great point!

sobolevn: 👍

thepabloaguilar:
> Awesome! Can you please also try to enable `mypy` plugin for it?
> Here: https://github.com/dry-python/returns/blob/master/returns/contrib/mypy/decorator_plugin.py#L34

Sure, I'll!!

PR #362, opened 2020-05-01
Adds `is_io` function

### LINKED ISSUE #309: Add `is_io` function
It should return `True` for `IO`, `IOResult`, and `RequiresContextIOResult`

thepabloaguilar:
I've implemented the function, wrote tests, wrote examples in documentation.
Unfortunately, I've imported `RequiresContextIOResult` class within `is_io` function to avoid circular import.

Closes #309


returns/functions.py
```diff
@@ -167,3 +170,20 @@ def not_(function: Callable[..., bool]) -> Callable[..., bool]:
     def wrapped_function(*args, **kwargs) -> bool:  # noqa: WPS430
         return not function(*args, **kwargs)
     return wrapped_function
+
+
+def is_io(container: BaseContainer) -> bool:
```
sobolevn, line 175: Is it possible to move it to `pipeline.py`? I am asking because I want to get rid of this nested import.

sobolevn: Nope, we should add a new module for it.
We would add `is_context, `is_result`, and `is_future`.

Let's call it `instances.py` maybe?

thepabloaguilar: What about `containers.py`? Because all of those functions will do something with containers!

thepabloaguilar: @sobolevn ?

sobolevn: Containers is too general. We check if something is an `instance` of `io`, `future`, etc. So, `instances.py` looks better for me.

thepabloaguilar: `instances` is too general, as well, this topic is much abstract kkkkk. But it's, I'll use `instances.py`!

sobolevn:
Can you please open new issue for the similar functions but for `future`, `context`, `result`?

thepabloaguilar:
> Can you please open new issue for the similar functions but for `future`, `context`, `result`?

Sure!!

PR #366, opened 2020-05-05
Add `is_result` function

### LINKED ISSUE #365: Add `is_result` function
It should return `True` for `Result`, `FutureResult`. `RequiresContextResult`, `ContextResult`, `RequiresContextIOResult` and `ContextIOResult`

thepabloaguilar:
I've implemented the function, wrote tests, wrote examples in documentation.

Closes #365

### DIFF
--- a/returns/instances.py
+++ b/returns/instances.py
@@ -1,6 +1,13 @@
-from returns.context import RequiresContextIOResult
+from returns.context import (
+    ContextIOResult,
+    ContextResult,
+    RequiresContextIOResult,
+    RequiresContextResult,
+)
+from returns.future import FutureResult
 from returns.io import IO, IOResult
 from returns.primitives.container import BaseContainer
+from returns.result import Result
 
 
 def is_io(container: BaseContainer) -> bool:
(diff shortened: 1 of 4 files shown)


thepabloaguilar:
@sobolevn Can you take a look, please?

returns/instances.py
```diff
@@ -17,3 +24,28 @@ def is_io(container: BaseContainer) -> bool:
...
+    result_containers = (
+        Result,
+        FutureResult,
+        ContextResult,
+        ContextIOResult,
+        IOResult,
+        RequiresContextResult,
+        RequiresContextIOResult,
+    )
+    return isinstance(container, result_containers)
```
amaksymov, line 51: I have some nit.
I think a better name for the variable is `result_container_type` than `result_containers` and in this case maybe the best not to use a temporary variable for one call and just use:
```
  return isinstance(container, (
    Result,
    FutureResult,
    ContextResult,
    ContextIOResult,
    IOResult,
    RequiresContextResult,
    RequiresContextIOResult,
  ))
```
Sorry, if this remark is meaningless.

thepabloaguilar: I agree with the variable's name, but not with removing the variable!
Using the variable it's more readable than passing all possible types to `isinstance` function.

thepabloaguilar:
I'll close this PT 'cause that functionality is not required so far!

PR #370, opened 2020-05-12
add @asyncize decorator

### LINKED ISSUE #367: Create `@asyncize` decorator
It should transform 

```python
@asyncize
def x(a: int) -> int: ...
```

Into:

```python
async def x(a: int) -> int: ...
```

It can be useful for working with `Future`s.

nsidnev:
Closes #367

### DIFF
--- a/returns/future.py
+++ b/returns/future.py
@@ -444,6 +444,9 @@ def asyncize(function: Callable[..., _ValueType]) -> Callable[
     """
     Decorator to turn a common function into an asynchronous function.
 
+    This decorator is useful for use with ``Future`` and ``FutureResult``
+    containers.
+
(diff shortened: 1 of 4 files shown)


returns/future.py
```diff
@@ -437,6 +437,31 @@ def decorator(*args, **kwargs):
     return decorator
 
 
+def asyncize(function: Callable[..., _ValueType]) -> Callable[
+    ...,
+    Coroutine[Any, Any, _ValueType],
```
sobolevn, line 442: Can you please add a note here that it is useful for composition with `Future` and `FutureResult`?

nsidnev: Done!

PR #396, opened 2020-06-01
Fix flake8 warnings in code blocks in docs

orsinium:
I'm preparing RST support for [flakehell](https://github.com/life4/flakehell/) and so happens that I tested in on Returns' docs. There are fixed violations (only pyflakes and pycodestyle for now). It's mostly about indentation. A bit nitpicky but a good start :)


orsinium:
Thank you :)

Issue #258, opened 2020-02-07
Consider adding validators package


sobolevn:
Currently a lot of python tools just throw exceptions at us or even uses: `.is_valid()` and `.errors` properties. It is not type-safe at all. Sometimes it is hard to explain to people how to really use type-driven design and make invalid states unreachable. 

My idea is that we need to write thin wrappers around popular typed validation packages: like `pydantic`, `cerebrus`, `schema`, etc to return really typed things.

We also need to think about API to be expressive and recognisable enough.

Related: https://gist.github.com/maksbotan/14b4bebda2acab98cdd158f85a970855
CC @maksbotan

sobolevn:
Shame on me! Not `validators` but `parsers`!

See https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

SaschaSchlemmer:
there is also [https://validators.readthedocs.io/en/latest/](https://validators.readthedocs.io/en/latest/) which imho fortunately returns errors instead of raising them

sobolevn:
1. https://hackage.haskell.org/package/monad-validate-1.2.0.0/docs/Control-Monad-Validate.html
2. https://github.com/debasishg/hask/blob/master/hfrdomain/README.md

sobolevn:
Or we can add `validate` function to the `pipeline.py`.
Probably in the next release.

sobolevn:
https://dry-rb.org/gems/dry-monads/1.3/validated/

sobolevn:
https://typelevel.org/cats/datatypes/validated.html

sobolevn:
https://typelevel.org/cats/datatypes/validated.html

sobolevn:
So, it looks like `Validated` should be defined like so:

```python
class Validated(Generic[_ValueType, _ErrorType]):
    _inner_value: Iterable[
        Callable[[_ValueType], Result[_ValueType, _ErrorType]],
    ]
```

It should work with both `IOResult` and `Result` somehow. `Union`?

sobolevn:
https://typelevel.org/cats/datatypes/validated.html#validated-vs-either

maksbotan:
Take a look at http://hackage.haskell.org/package/these as well.

http://hackage.haskell.org/package/these-1.1/docs/Data-These.html#t:These
http://hackage.haskell.org/package/monad-chronicle-1/docs/Control-Monad-Chronicle.html

[...]

PR #422, opened 2020-06-13
Implements `trace` for failures

### LINKED ISSUE #409: Improve Failure tracibility
`dry-rb` has a cool feature: tracing failures https://dry-rb.org/gems/dry-monads/1.3/tracing-failures/
We need something similar to improve debugging experience. It might be included into the core, or just into some `contrib/` package.

thepabloaguilar:
All context can be found on #409 issue!

Just a little explanation about `_get_trace` **function**:
```python
def _get_trace(_self) -> List[FrameInfo]:
    current_stack = stack()
    return current_stack[2:]
```
The return is the current call stack starting from the third position, to avoid two non useful calls on the call stack and you can see in the example below what calls they're:

**Complete call stack**
```text
/home/pabloaguilar/Projects/returns/returns/tracing.py:41 in `_get_trace`
/home/pabloaguilar/Projects/returns/returns/result.py:376 in `__init__`
/home/pabloaguilar/Projects/returns/returns/result.py:521 in `Failure`
/home/pabloaguilar/Projects/returns/example.py:33 in `my_func_one`
/home/pabloaguilar/Projects/returns/example.py:67 in `<module>`
```
**call stack starting from third position**
```text
/home/pabloaguilar/Projects/returns/returns/result.py:521 in `Failure`
/home/pabloaguilar/Projects/returns/example.py:33 in `my_func_one`
/home/pabloaguilar/Projects/returns/example.py:67 in `<module>`
```

closes #409


thepabloaguilar:
Should I have add something to `CHANGELOG.md`?

returns/tracing.py
```diff
@@ -0,0 +1,42 @@
...
+    setattr(_Failure, '_get_trace', substitute_get_trace)  # noqa: B010
+    try:
+        yield
+    finally:
+        setattr(_Failure, '_get_trace', unpatched_get_trace)  # noqa: B010
+
+
+def _get_trace(_self) -> List[FrameInfo]:
+    current_stack = stack()
+    return current_stack[2:]
```
thepabloaguilar, line 42: Should I have change the signature of `_get_trace` **function** to match with `_get_trace` **method**?
```suggestion
def _get_trace(_self) -> Optional[List[FrameInfo]]:
    current_stack = stack()
    return current_stack[2:]
```

docs/pages/tracing.rst
```diff
@@ -0,0 +1,63 @@
+Tracing
```
sobolevn, line 1: This should be listed under `Integration`, let's rename it to "Development"

docs/pages/tracing.rst
```diff
@@ -0,0 +1,63 @@
...
+=======
+
+collect_traces
```
sobolevn, line 4: Let's name it `Tracing Failures`

docs/pages/tracing.rst
```diff
@@ -0,0 +1,63 @@
...
+  >>> non_traced_failure = Failure('Normal Failure')
+  >>> traced_failure = traced_function('Traced Failure')
+
+  >>> assert non_traced_failure.trace is None
+  >>> assert isinstance(traced_failure.trace, list)
+  >>> assert isinstance(traced_failure.trace[0], FrameInfo)
+
+.. warning::
+
+  Activating trace can make your program a bit slower if it has many points where ``Failure`` is often created.
```
sobolevn, line 57: We need to add a warning about thread safety. I guess it is not safe to use in multiple threads. It might have side effects. Am I right?

returns/primitives/container.py
```diff
@@ -12,16 +13,18 @@
 class BaseContainer(Immutable, metaclass=ABCMeta):
     """Utility class to provide all needed magic methods to the context."""
 
-    __slots__ = ('_inner_value',)
+    __slots__ = ('_inner_value', '_trace')
```
sobolevn, line 16: Not all containers have `_trace` 

Can we only change this in `Result`?

returns/tracing.py
```diff
@@ -0,0 +1,42 @@
...
+        >>> from returns.io import IOResult
+        >>> from returns.result import Result
+        >>> from returns.tracing import collect_traces
+
+        >>> with collect_traces():
+        ...     traced_failure = Result.from_failure('Traced Failure')
+        >>> non_traced_failure = IOResult.from_failure('Non Traced Failure')
+
+        >>> assert non_traced_failure.trace is None
+        >>> assert isinstance(traced_failure.trace, list)
```
sobolevn, line 27: Can we make this value visual? Let's not use `assert` here and cast it to text format.
This way user would know what to expect.

returns/tracing.py
```diff
@@ -0,0 +1,42 @@
...
+    unpatched_get_trace = getattr(_Failure, '_get_trace')  # noqa: B009
+    substitute_get_trace = types.MethodType(_get_trace, _Failure)
+    setattr(_Failure, '_get_trace', substitute_get_trace)  # noqa: B010
+    try:
+        yield
+    finally:
+        setattr(_Failure, '_get_trace', unpatched_get_trace)  # noqa: B010
+
+
+def _get_trace(_self) -> Optional[List[FrameInfo]]:
```
sobolevn, line 40: Let's document this value as well.

docs/pages/tracing.rst
thepabloaguilar: `Tracing -> Development` or `Integration  -> Development`?

returns/tracing.py
thepabloaguilar: `_self`?

sobolevn: The whole function.

returns/primitives/container.py
thepabloaguilar: Done!

docs/pages/tracing.rst
thepabloaguilar: Done!

docs/pages/tracing.rst
thepabloaguilar: Done!

returns/tracing.py
thepabloaguilar: Could you verify it, please?

returns/tracing.py
thepabloaguilar: One discussion we didn't have about **trace** property, should it return the list of *FrameInfo* or print out the list properly formated?

docs/pages/tracing.rst
thepabloaguilar: I think you're right. But I'll make a test to document this fact!

thepabloaguilar:
```python
from time import sleep
from typing import Any

from returns.result import Failure
from returns.tracing import collect_traces


def my_func_one(have_to_patch: bool) -> Any:
    if have_to_patch:
        with collect_traces():
            print(f'have_to_patch: {have_to_patch} | sleep')
            sleep(10)
            print(f'Exiting Monkey Patch')
    else:
        print(f'have_to_patch: {have_to_patch} | sleep')
        sleep(5)
        print(f'have_to_patch: {have_to_patch} | trace: {Failure("some value").trace}')
        sleep(6)
        print(f'have_to_patch: {have_to_patch} | trace: {Failure("some value").trace}')


if __name__ == "__main__":
    from concurrent import futures

    with futures.ThreadPoolExecutor(max_workers=2) as executor:
        executor.map(my_func_one, [True, False])
```

**Output**
```text
have_to_patch: True | sleep
have_to_patch: False | sleep
have_to_patch: False | trace: [ OMITTED TRACE ]
Exiting Monkey Patch
have_to_patch: False | trace: None
```

thepabloaguilar: I'm going to write the warning now!

returns/tracing.py
sobolevn: It should return a `list`, this more like FP 🙂

returns/tracing.py
```diff
@@ -0,0 +1,63 @@
...
+
+        >>> with collect_traces():
+        ...     traced_failure = Result.from_failure('Traced Failure')
+        >>> non_traced_failure = IOResult.from_failure('Non Traced Failure')
+
+        >>> assert non_traced_failure.trace is None
+        >>> assert isinstance(traced_failure.trace, list)
+        >>> assert isinstance(traced_failure.trace[0], FrameInfo)
+        >>> traced_failure.trace  # doctest: +SKIP
+        [FrameInfo(frame=<frame at 0x7f, file '.../returns/returns/result.py', line 525, code Failure>, ...]
```
sobolevn, line 30: We can prettify it somehow. Our goal it to make it readable. Possibly `pprint`?

docs/pages/development.rst
```diff
@@ -0,0 +1,68 @@
...
+  >>> assert isinstance(traced_failure.trace, list)
+  >>> assert isinstance(traced_failure.trace[0], FrameInfo)
+
+.. warning::
+
+  Activating trace can make your program a bit slower if it has many points where ``Failure`` is often created.
+
+.. warning::
+
+  ``collect_traces`` is not Thread Safety, beware to use it!
```
sobolevn, line 62:
```
``collect_traces`` is not thread safe, beware to use it!
```

docs/pages/development.rst
```diff
@@ -0,0 +1,68 @@
...
+  >>> non_traced_failure = Failure('Normal Failure')
+  >>> traced_failure = traced_function('Traced Failure')
+
+  >>> assert non_traced_failure.trace is None
+  >>> assert isinstance(traced_failure.trace, list)
+  >>> assert isinstance(traced_failure.trace[0], FrameInfo)
+
+.. warning::
+
+  Activating trace can make your program a bit slower if it has many points where ``Failure`` is often created.
```
sobolevn, line 58: We need one more warning:

```
Traces are meant to use during development only. 
```

docs/pages/development.rst
```diff
@@ -0,0 +1,68 @@
...
+  >>> from inspect import FrameInfo
+
+  >>> from returns.result import Failure
+  >>> from returns.tracing import collect_traces
+
+  >>> non_traced_failure = Failure('Normal Failure')
+  >>> with collect_traces():
+  ...     traced_failure = Failure('Traced Failure')
+
+  >>> assert non_traced_failure.trace is None
```
sobolevn, line 31: The same here, we need to make traces visual.

docs/pages/development.rst
```diff
@@ -0,0 +1,68 @@
...
+
+  >>> @collect_traces()
+  ... def traced_function(value: str) -> IOResult[str, str]:
+  ...     return IOFailure(value)
+
+  >>> non_traced_failure = Failure('Normal Failure')
+  >>> traced_failure = traced_function('Traced Failure')
+
+  >>> assert non_traced_failure.trace is None
+  >>> assert isinstance(traced_failure.trace, list)
```
sobolevn, line 53: And here.

returns/tracing.py
thepabloaguilar: Maybe we can add a *for loop* with *print*!
```python
>>> for trace_line in traced_failure.trace:
...     print(f"{trace_line.filename}:{trace_line.lineno} in `{trace_line.function}`")
...
/returns/returns/result.py:529 in `Failure`
/example_folder/example.py:26 in `my_example_func`
/example_folder/example.py:53 in `<module>`
```
What do you think?

returns/tracing.py
sobolevn: Looks good!

returns/tracing.py
sobolevn: ++

thepabloaguilar:
> Awesome!
> 
> I also got an idea that we should add an opportunity to test traces with out `pytest` plugins.
> This should help people to predict which paths the code will take.
> 
> But, I am not sure about an API. Do you have any ideas?

I didn't understand your idea, sorry!
Could you provide an example or more details?

sobolevn:
> I didn't understand your idea, sorry!
> Could you provide an example or more details?

I will create a new issue for it.

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
+Development
+===========
+
+Tracing Failures
+----------------
+
+Sometimes we want to trace where the ``Failure`` has occurred in our system,
+`returns` provide a way to trace those failures. By default tracing is disabled.
```
sobolevn, line 8:
```
`returns`
```

to 

```
``returns``
```

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
...
+
+The trace is accessible by **trace** property that is available for ``Result``,
```
sobolevn, line 10:
```
**trace**
```

to

```
:meth:`trace <returns.result.Result.trace>`
```

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
...
+``IOResult`` containers. It's basically a list containing all
+:class:`inspect.FrameInfo` from the call stack when the ``Failure`` was created.
```
sobolevn, line 12:
```
all :class:`inspect.FrameInfo` objects
```

sobolevn: `was originally created.`

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
...
+
+To enable it you can use
+:func:`collect_traces <returns.tracing.collect_traces>`.
+See some examples bellow:
```
sobolevn, line 16: `See examples bellow.`

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
...
+
+  >>> non_traced_failure = get_failure('Normal Failure')
+  >>> with collect_traces():
+  ...     traced_failure = get_failure('Traced Failure')
+
+  >>> assert non_traced_failure.trace is None
+  >>> assert isinstance(traced_failure.trace, list)
+  >>> assert isinstance(traced_failure.trace[0], FrameInfo)
+
+  >>> str(non_traced_failure.trace)
```
sobolevn, line 38: This can be removed, `assert is None` is enough

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
...
+  ...     return IOFailure(value)
+
+  >>> non_traced_failure = Failure('Normal Failure')
+  >>> traced_failure = traced_function('Traced Failure')
+
+  >>> assert non_traced_failure.trace is None
+  >>> assert isinstance(traced_failure.trace, list)
+  >>> assert isinstance(traced_failure.trace[0], FrameInfo)
+
+  >>> str(non_traced_failure.trace)
```
sobolevn, line 68: The same here

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
...
+  >>> @collect_traces()
+  ... def traced_function(value: str) -> IOResult[str, str]:
+  ...     return IOFailure(value)
+
+  >>> non_traced_failure = Failure('Normal Failure')
+  >>> traced_failure = traced_function('Traced Failure')
+
+  >>> assert non_traced_failure.trace is None
+  >>> assert isinstance(traced_failure.trace, list)
+  >>> assert isinstance(traced_failure.trace[0], FrameInfo)
```
sobolevn, line 66: Let's move this to `for` loop.

```python
assert isinstance(trace_line, FrameInfo)
```

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
...
+  ...
+  /returns/returns/result.py:525 in `Failure`
+  /returns/returns/io.py:852 in `IOFailure`
+  /example_folder/example.py:7: in `traced_function`
+  /usr/lib/python3.8/contextlib.py:75 in `inner`
+  /example_folder/example.py:1 in `<module>`
+
+.. warning::
+
+  Activating trace can make your program a bit slower if it has many points where ``Failure`` is often created.
```
sobolevn, line 81: `Activating trace can make your program noticeably slower`

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
...
+
+.. warning::
+
+  ``collect_traces`` is not thread safe, beware to use it!
```
sobolevn, line 85: `beware to use it with threading!`

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
...
+
+.. warning::
+
+  Traces are meant to use during development only.
```
sobolevn, line 89: `to be used`

docs/pages/development.rst
```diff
@@ -0,0 +1,95 @@
...
+
+API Reference
+-------------
+
+.. automodule:: returns.tracing
```
sobolevn, line 94: Let's move it into https://github.com/dry-python/returns/tree/master/returns/primitives

Because it does not have the same level of importance as core features.

returns/result.py
```diff
@@ -47,14 +50,22 @@ class Result(
 
     """
 
+    __slots__ = ('_trace',)
```
sobolevn, line 53: Are `__slots__` combined when inheritance is used? (sorry, I don't know how it works)

I am asking because I was expecting to see  `__slots__ = ('_inner_value', '_trace',)`

returns/tracing.py
```diff
@@ -0,0 +1,69 @@
...
+
+        >>> with collect_traces():
+        ...     traced_failure = Result.from_failure('Traced Failure')
+        >>> non_traced_failure = IOResult.from_failure('Non Traced Failure')
+
+        >>> assert non_traced_failure.trace is None
+        >>> assert isinstance(traced_failure.trace, list)
+        >>> assert isinstance(traced_failure.trace[0], FrameInfo)
+
+        >>> str(non_traced_failure.trace)
```
sobolevn, line 30: The same about `None` and `for` loop here.

returns/tracing.py
```diff
@@ -0,0 +1,69 @@
...
+
+    This function is the substitute for '_get_trace' method from ``_Failure``
+    class on Monkey Patching promoted by :func:`returns.tracing.collect_traces`
+    function.
+
+    We get all the call stack from the current call and return it from the
+    third position, to avoid two non-useful calls on the call stack.
+    Those non-useful calls are a call to this function and a call to `__init__`
+    method from ``_Failure`` class. We're just interested in the call stack
+    ending on `Failure` function call!
```
sobolevn, line 62:
```
`Failure`
```

to 

```
``Failure``
```

returns/result.py
thepabloaguilar: I had the same question! Yes, `__slots__` are combined and if we overlap an attribute in child class it can waste memory!
[How does inheritance of \_\_slots__ in subclasses actually work?](https://stackoverflow.com/questions/1816483/how-does-inheritance-of-slots-in-subclasses-actually-work)

Another point here is overlapping slots attributes can be tricky, take a look at the sixth item:
[Notes on using slots](https://docs.python.org/3/reference/datamodel.html#notes-on-using-slots)

docs/pages/development.rst
thepabloaguilar: I've chosen to keep that assert out of `for` loop because of the entire for loop is skiped because the `print` statement. I've modified the assert:
```python
assert all(isinstance(trace_line, FrameInfo) for trace_line in traced_failure.trace)
```

thepabloaguilar:
> Looks like almost done!
> 
> I am pretty sure that this is the last round of review and we are going to merge this soon.
> You have my gratitude, thanks a lot!

Thanks man, it's a pleasure to contribute to `returns`!

returns/result.py
sobolevn: TIL 👍

sobolevn:
Thanks again! 

P.S. Can you please take a look at #258? I am researching possible design ideas right now. I would love to hear your thoughts.

thepabloaguilar:
> P.S. Can you please take a look at #258? I am researching possible design ideas right now. I would love to hear your thoughts.

Yes, I'll take a look asap!

PR #442, opened 2020-06-27
Adds missing return type to `collect_traces` function

### LINKED ISSUE #441: collect_traces() is missing a return type
It should be `Iterator[None]`.
We should also add a typesafety test. 

@thepabloaguilar would you please do it? 🙂

### LINKED ISSUE #443: `collect_traces` wrong return type when it's used as a decorator
We're using the `@context_manager` decorator on `collect_traces` function, so, we can use it as a decorator or context manager.
When we use it as a decorator, the decorated function be with a wrong return type!!

Type safety test:
```yml
- case: collect_traces_context_manager_return_type
  disable_cache: true
  main: |
    from returns.primitives.tracing import collect_traces

    @collect_traces
    def function() -> int:
      return 0

    reveal_type(function)  # N: Revealed type is 'def () -> builtins.int'
```

Expected:
```text
Revealed type is 'def () -> builtins.int'
```

Actual:
```text
Revealed type is 'contextlib._GeneratorContextManager[None]'
```

Refs #441

thepabloaguilar:
closes #441
closes #443

### DIFF
--- a/returns/primitives/tracing.py
+++ b/returns/primitives/tracing.py
@@ -14,7 +14,7 @@ from typing import (
 
 from returns.result import _Failure
 
-Function = TypeVar('Function', bound=Callable)
+_FunctionType = TypeVar('_FunctionType', bound=Callable)
 
 
 @overload
@@ -23,13 +23,13 @@ def collect_traces() -> ContextManager[None]:
 
 
 @overload
-def collect_traces(function: Function) -> Function:
+def collect_traces(function: _FunctionType) -> _FunctionType:
     """Decorator to active traces collect to the Failures."""
 
 
 def collect_traces(
-    function: Optional[Function] = None,
-) -> Union[Function, ContextManager[None]]:  # noqa: DAR101, DAR201, DAR301
+    function: Optional[_FunctionType] = None,
+) -> Union[_FunctionType, ContextManager[None]]:  # noqa: DAR101, DAR201, DAR301
     """
     Context Manager/Decorator to active traces collect to the Failures.
 
--- a/returns/primitives/tracing.py
+++ b/returns/primitives/tracing.py
@@ -1,12 +1,35 @@
 import types
 from contextlib import contextmanager
 from inspect import FrameInfo, stack
-from typing import Iterator, List, Optional
+from typing import (
+    Callable,
+    ContextManager,
+    Iterator,
+    List,
+    Optional,
+    TypeVar,
+    Union,
+    overload,
+)
 
 from returns.result import _Failure
 
+Function = TypeVar('Function', bound=Callable)
 
-def collect_traces() -> Iterator[None]:
+
+@overload
+def collect_traces() -> ContextManager[None]:
+    """Context Manager to active traces collect to the Failures."""
+
+
+@overload
+def collect_traces(function: Function) -> Function:
+    """Decorator to active traces collect to the Failures."""
+
+
+def collect_traces(
+    function: Optional[Function] = None,
+) -> Union[Function, ContextManager[None]]:  # noqa: DAR101, DAR201, DAR301
     """
     Context Manager/Decorator to active traces collect to the Failures.
 
@@ -35,13 +58,16 @@ def collect_traces() -> Iterator[None]:
         # doctest: # noqa: DAR301, E501
 
     """
-    unpatched_get_trace = getattr(_Failure, '_get_trace')  # noqa: B009
-    substitute_get_trace = types.MethodType(_get_trace, _Failure)
-    setattr(_Failure, '_get_trace', substitute_get_trace)  # noqa: B010
-    try:  # noqa: WPS501
-        yield
-    finally:
-        setattr(_Failure, '_get_trace', unpatched_get_trace)  # noqa: B010
+    def factory() -> Iterator[None]:
+        unpatched_get_trace = getattr(_Failure, '_get_trace')  # noqa: B009
+        substitute_get_trace = types.MethodType(_get_trace, _Failure)
+        setattr(_Failure, '_get_trace', substitute_get_trace)  # noqa: B010
+        try:  # noqa: WPS501
+            yield
+        finally:
+            setattr(_Failure, '_get_trace', unpatched_get_trace)  # noqa: B010
+    collect_traces_cm = contextmanager(factory)()
+    return collect_traces_cm(function) if function else collect_traces_cm
 
 
 def _get_trace(_self: _Failure) -> Optional[List[FrameInfo]]:
@@ -64,6 +90,3 @@ def _get_trace(_self: _Failure) -> Optional[List[FrameInfo]]:
     """
     current_stack = stack()
     return current_stack[2:]
-
-
-collect_traces = contextmanager(collect_traces)  # type: ignore
--- a/returns/primitives/tracing.py
+++ b/returns/primitives/tracing.py
@@ -58,6 +58,7 @@ def collect_traces(
         # doctest: # noqa: DAR301, E501
 
     """
+    @contextmanager
     def factory() -> Iterator[None]:
         unpatched_get_trace = getattr(_Failure, '_get_trace')  # noqa: B009
         substitute_get_trace = types.MethodType(_get_trace, _Failure)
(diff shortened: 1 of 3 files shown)


typesafety/test_primitives/test_tracing/test_collect_traces.yml
```diff
@@ -0,0 +1,13 @@
...
+    from returns.primitives.tracing import collect_traces
+
+    reveal_type(collect_traces)  # N: Revealed type is 'def () -> contextlib._GeneratorContextManager[None]'
+
+- case: collect_traces_context_manager_return_type
+  disable_cache: true
+  main: |
+    from returns.primitives.tracing import collect_traces
+
+    reveal_type(collect_traces())  # N: Revealed type is 'contextlib._GeneratorContextManager[None]'
```
sobolevn, line 13: Can we use `@overload` to fix this type?

thepabloaguilar: I've tried without success!

Implementation:
```python
@overload
def collect_traces() -> Iterator[None]:
    ...

@contextmanager
def collect_traces():  # I've tried with `def collect_traces() -> Iterator[None]` as well
    # IMPLEMENTATION
```

Errors:
```sh
/returns/returns/primitives/tracing:9: error: Single overload definition, multiple required
/returns/returns/primitives/tracing:14: error: Overloaded function implementation cannot produce return type of signature 1
```

thepabloaguilar: I've found a possible answer, https://github.com/python/mypy/issues/1317#issuecomment-207564973

```python
def collect_traces() -> Iterator[None]:
    # IMPLEMENTATION

collect_traces = contextmanager(collect_traces)  # type: ignore
```

Output of `reveal_type`:
```sh
Revealed type is 'def () -> typing.Iterator[None]'
```

I didn't like this resolution so much but it works!
What do you think?

thepabloaguilar: The above solution fixes #443, as well!

sobolevn: Awesome! 👍

thepabloaguilar: I was wrong about this:
> The above solution fixes #443, as well!

The solution cannot fixes, see the errors below:
```sh
E   pytest_mypy_plugins.utils.TypecheckAssertionError: Invalid output: 
E   Expected:
E     main:7: note: Revealed type is 'def () -> builtins.int' (diff)
E   Actual:
E     main:3: error: Too many arguments for "collect_traces" (diff)
E     main:7: note: Revealed type is 'typing.Iterator[None]' (diff)
E   
E   Alignment of first line difference:
E     E: main:7: note: Revealed type is 'def () -> builtins.int'
E     A: main:3: error: Too many arguments for "collect_traces"
```

But fixes the current issue about return type when used as a context manager!

typesafety/test_primitives/test_tracing/test_collect_traces.yml
```diff
@@ -0,0 +1,13 @@
+- case: collect_traces_context_manager_return_type_hint
+  disable_cache: true
+  main: |
+    from returns.primitives.tracing import collect_traces
+
+    reveal_type(collect_traces)  # N: Revealed type is 'def () -> typing.Iterator[None]'
+
+- case: collect_traces_context_manager_return_type
```
sobolevn, line 8: We also need to test `with collect_traces():` here.

returns/primitives/tracing.py
```diff
@@ -1,13 +1,12 @@
...
 from inspect import FrameInfo, stack
-from typing import List, Optional
+from typing import Iterator, List, Optional
 
 from returns.result import _Failure
 
 
-@contextmanager
-def collect_traces():
+def collect_traces() -> Iterator[None]:
```
sobolevn, line 9: Will this work? 

```python
from contextlib import contextmanager
from typing import overload, Iterator, Callable, TypeVar, TYPE_CHECKING, ContextManager, Optional, Union

if not TYPE_CHECKING:
    reveal_type = print

F = TypeVar('F', bound=Callable)


@overload
def collect_traces(f: F) -> F:
    ...

@overload
def collect_traces() -> ContextManager[None]:
    ...

def collect_traces(f: Optional[F] = None) -> Union[F, ContextManager[None]]:
    @contextmanager
    def factory() -> Iterator[None]:
        try:
            yield
        finally:
            print('done')
    return f if f is not None else factory()

@collect_traces
def some(a: int) -> str:
    return str(a)

with collect_traces():
    ...

reveal_type(some(1))
```

mypy is happy about it:

```
» mypy ex.py
ex.py:34: note: Revealed type is 'builtins.str'
```

typesafety/test_primitives/test_tracing/test_collect_traces.yml
sobolevn: And `@collect_traces` decorator.

returns/primitives/tracing.py
thepabloaguilar: I've updated the implementation using your suggestion:

```python
def collect_traces(
    function: Optional[Function] = None,
) -> Union[Function, ContextManager[None]]:
    def factory() -> Iterator[None]:
        unpatched_get_trace = getattr(_Failure, '_get_trace')
        substitute_get_trace = types.MethodType(_get_trace, _Failure)
        setattr(_Failure, '_get_trace', substitute_get_trace)
        try:
            yield
        finally:
            setattr(_Failure, '_get_trace', unpatched_get_trace)
    collect_traces_cm = contextmanager(factory)()
    return (
        collect_traces_cm(function) if function else collect_traces_cm
    )
```

Some explanations:
```python
collect_traces_cm = contextmanager(factory)()
```
To avoid mypy error I've decided to pass the factory function directly to `contextmanager` function instead of using it as a decorator

---

```python
return (
    collect_traces_cm(function) if function else collect_traces_cm
)
```
`contextmanager(factory)()` returns an instance of `_GeneratorContextManager` class that implements `__enter__`, `__exit__` and `__call__`. This means to use like a common context manager we just return it, but to use it like decorator we have to pass trough the callable instance!

returns/primitives/tracing.py
```diff
@@ -1,13 +1,35 @@
...
+    List,
+    Optional,
+    TypeVar,
+    Union,
+    overload,
+)
 
 from returns.result import _Failure
 
+Function = TypeVar('Function', bound=Callable)
```
sobolevn, line 17: This should be named `_FunctionType`
1. It is protected
2. We try to use `Type` suffix to indicate that this is a typevar

typesafety/test_primitives/test_tracing/test_collect_traces.yml
```diff
@@ -0,0 +1,35 @@
...
+    reveal_type(collect_traces)  # N: Revealed type is 'Overload(def () -> typing.ContextManager[None], def [Function <: def (*Any, **Any) -> Any] (function: Function`-1) -> Function`-1)'
+
+- case: collect_traces_context_manager_return_type_two
+  disable_cache: true
+  main: |
+    from returns.primitives.tracing import collect_traces
+
+    reveal_type(collect_traces())  # N: Revealed type is 'typing.ContextManager[None]'
+
+- case: collect_traces_decorated_function_return_type
```
sobolevn, line 15: We can safely replace this test with `with collect_traces()` one.

returns/primitives/tracing.py
thepabloaguilar: Done!

typesafety/test_primitives/test_tracing/test_collect_traces.yml
thepabloaguilar: I don't know if I have did right:
```python
with reveal_type(collect_traces()):  # N: Revealed type is 'typing.ContextManager[None]'
    pass
```

thepabloaguilar:
I was wrong in this explanation, in fact is just a PyCharm error/warning!
> ```python
> collect_traces_cm = contextmanager(factory)()
> ```
> 
> To avoid mypy error I've decided to pass the factory function directly to `contextmanager` function instead of using it as a decorator

I've made the modification to use `contextmanager` as a decorator!

sobolevn:
Thanks a lot for the great work!

PR #446, opened 2020-07-04
WIP: adds hkt support

sobolevn:
This is just a quick demo for ones who is interested.


sobolevn:
The only thing left to do is:
- add `Rescuable` interface with `.rescue` method
- add `rescue` method and pointfree function
- add `ResultBased` interface with `.from_result` and `.bind_result` methods
- add `bind_result` method and pointfree function

PR #455, opened 2020-07-07
Fix documentation to compose error types

### LINKED ISSUE #454: Example error in `Result` container section
We have this example in "[How to compose error types?](https://returns.readthedocs.io/en/latest/pages/result.html#how-to-compose-error-types)" section:

```python
>>> from returns.result import Result, Success

>>> def div(number: int) -> Result[float, ZeroDivisionError]:
...     return Success(1 / number)

>>> container: Result[int, ValueError] = Success(1)
>>> str(container.unify(div))
'<Success: 1.0>'

>>> # => Revealed type is:
>>> # Result[float, Union[ValueError, ZeroDivisionError]]
```

`div` function is not safe if we pass `0 (zero)` as an argument!
I know that example is about `unify` method, but we can make a little improvement there, maybe:
```python
def div(number: int) -> Result[float, ZeroDivisionError]:
    if number:
        return Success(1 / number)
    return Failure(ZeroDivisionError('division by zero'))
```

I'm opening this issue because a friend of mine had some questions around `div`.

kenjihiraoka:
This project is very nice and meaningful! I hope that I can understand more and more how to use it. :smile: 

closes #454

### DIFF
--- a/docs/pages/result.rst
+++ b/docs/pages/result.rst
@@ -170,10 +170,12 @@ Like so:
 
 .. code:: python
 
-  >>> from returns.result import Result, Success
+  >>> from returns.result import Result, Success, Failure
 
   >>> def div(number: int) -> Result[float, ZeroDivisionError]:
-  ...     return Success(1 / number)
+  ...     if number:
+  ...         return Success(1 / number)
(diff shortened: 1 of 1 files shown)


sobolevn:
Thanks a lot!

Issue #457, opened 2020-07-08
Test monad laws with a custom pytest plugin


sobolevn:
1. https://arrow-kt.io/docs/0.10/arrow/typeclasses/monad/#laws
2. https://github.com/arrow-kt/arrow-core/blob/master/arrow-core/arrow-test/src/main/kotlin/arrow/test/laws/MonadLaws.kt

thepabloaguilar:
[Verifying Typeclass Laws in Haskell with QuickCheck](https://austinrochford.com/posts/2014-05-27-quickcheck-laws.html)

sobolevn:
Or, we can use data-types to represent different `Laws`: https://github.com/zio/zio/blob/f963adee37635f021a741d9d5ae4ca2ed4890b10/test/shared/src/main/scala/zio/test/laws/ZLawful.scala

For example:

```python
class Functor(Law[Laws.Identity], ...): ...
```

It looks like something I really want to research.

sobolevn:
Also:
- https://github.com/zio/zio/blob/f36ba93af21f4cdcd880a39bde7911f127ff1fc6/test/shared/src/main/scala/zio/test/laws/package.scala
- https://github.com/zio/zio/blob/f963adee37635f021a741d9d5ae4ca2ed4890b10/test-tests/shared/src/test/scala/zio/test/laws/LawsSpec.scala

This is super cool idea!

We can also add `hypothesis` as `returns[test]` dependency and use API like `returns.check_laws(my_custom_container)` in tests to test the hell out of it.

sobolevn:
Done in #533

Issue #451, opened 2020-07-07
Improve docs: why do we ignore almost all the FP jargon in this project


sobolevn:
- [ ] Describe our decision
- [ ] Add a link to: https://github.com/jmesyou/functional-programming-jargon.py
- [ ] Improve https://github.com/jmesyou/functional-programming-jargon.py with `returns` links and examples

thepabloaguilar:
Can FP users understand what is "a thing" without jargon? Can they correlate "a thing" to a jargon?
If the answers to these questions are "yes" I think we can "ignore" the FP jargon to be more inclusive with people that do not understand it!

*On the other hand, we have to get these people familiar with FP jargon to understand what they are using.*

thepabloaguilar:
What's the final decision here? Will we get out of FP jargons?

sobolevn:
Yes, we don't use it. (As a matter of fact, we already don't use it).

The history behind this is really interesting. The first name of this project was `dry-monads`: https://github.com/dry-python/returns/tree/0.1.0

But, then I got a lot of uneducated feedback. It was hard to manage. So, instead we went to make this concept hidden.
And it worked! 

Basically, we came up with our own terms that do mean the similar things. And we should stick to them.

thepabloaguilar:
So, the first item was completed I guess.
> * [X]  Describe our decision

Answer:

> then I got a lot of uneducated feedback. It was hard to manage. So, instead we went to make this concept hidden.
And it worked!
> Basically, we came up with our own terms that do mean the similar things. And we should stick to them.

---

Where we have to add the link? On the first page?
> * [ ]  Add a link to: https://github.com/jmesyou/functional-programming-jargon.py

What exactly we have to improve? 
> * [ ]  Improve https://github.com/jmesyou/functional-programming-jargon.py with `returns` links and examples

sobolevn:
> What exactly we have to improve?

There are a lot of concepts that are poorly illustrated in the original article: https://github.com/jmesyou/functional-programming-jargon.py 

I guess we can improve almost every point in there 🙂

thepabloaguilar:
If we create a repository inside `dry-python` it's easier to update it, you've opened a PR there 23 days ago and so far you haven't got any response!

sobolevn:
Great idea!

sobolevn:
New repo: https://github.com/dry-python/functional-jargon-python

thepabloaguilar:
How do you want to do this??
[...]

PR #470, opened 2020-07-13
Adds documentation for `Mappable` and `Bindable` interfaces

thepabloaguilar:
refs #517

### DIFF
--- a/docs/pages/interfaces.rst
+++ b/docs/pages/interfaces.rst
@@ -48,11 +48,11 @@ that implements the ``Mappable`` interface:
 
   >>> from returns.maybe import Maybe, Some
 
-  >>> def mappable(string: str) -> str:
+  >>> def can_be_mapped(string: str) -> str:
   ...     return string + '!'
 
   >>> maybe_str: Maybe[str] = Some('example')
-  >>> assert maybe_str.map(mappable) == Some('example!')
+  >>> assert maybe_str.map(can_be_mapped) == Some('example!')
 
 :class:`Mappable <returns.interfaces.mappable.MappableN>` interface help us to
 create our own mappable container like :class:`Maybe <returns.maybe.Maybe>`.
@@ -62,13 +62,13 @@ create our own mappable container like :class:`Maybe <returns.maybe.Maybe>`.
   >>> from typing import Any, Callable, TypeVar
 
   >>> from returns.interfaces.mappable import Mappable1
-  >>> from returns.primitives.hkt import Kind1
+  >>> from returns.primitives.hkt import SupportsKind1
 
   >>> _NumberType = TypeVar('_NumberType')
   >>> _NewNumberType = TypeVar('_NewNumberType')
 
   >>> class Number(
-  ...     Kind1['Number', _NumberType],
+  ...     SupportsKind1['Number', _NumberType],
   ...     Mappable1[_NumberType],
   ... ):
   ...     def __init__(self, inner_value: _NumberType):
@@ -111,11 +111,13 @@ the mappable has to be the same, unaltered.
   >>> mappable_number: Number[int] = Number(1)
   >>> assert mappable_number.map(identity) == Number(1)
 
-2. **Associative Law**: Given two functions, `x` and `y`, calling the map
-method with `x` function and after that calling with `y` function must have the
+2. **Associative Law**: Given two functions, ``x`` and ``y``, calling the map
+method with ``x`` function and after that calling with ``y`` function must have the
 same result if we compose them together.
 
 .. code:: python
+
+  >>> from returns.functions import compose
 
   >>> def add_one(number: int) -> int:
   ...     return number + 1
@@ -124,7 +126,7 @@ same result if we compose them together.
   ...     return number * 10
 
   >>> mappable_number: Number[int] = Number(9)
-  >>> assert mappable_number.map(add_one).map(multiply_by_ten) == mappable_number.map(lambda value: multiply_by_ten(add_one(value)))
+  >>> assert mappable_number.map(add_one).map(multiply_by_ten) == mappable_number.map(compose(add_one, multiply_by_ten))
 
 Bindable
 --------
--- a/docs/pages/interfaces.rst
+++ b/docs/pages/interfaces.rst
@@ -51,7 +51,7 @@ that implements the ``Mappable`` interface:
   >>> def mappable(string: str) -> str:
   ...     return string + '!'
 
-  >>> maybe_str: Maybe[str] = Maybe.from_value('example')
+  >>> maybe_str: Maybe[str] = Some('example')
   >>> assert maybe_str.map(mappable) == Some('example!')
 
 :class:`Mappable <returns.interfaces.mappable.MappableN>` interface help us to
@@ -59,14 +59,18 @@ create our own mappable container like :class:`Maybe <returns.maybe.Maybe>`.
 
 .. code:: python
 
-  >>> from typing import Any, Callable, TypeVar, Union
+  >>> from typing import Any, Callable, TypeVar
 
   >>> from returns.interfaces.mappable import Mappable1
+  >>> from returns.primitives.hkt import Kind1
 
-  >>> _NumberType = TypeVar('_NumberType', bound=Union[int, float, complex])
-  >>> _NewNumberType = TypeVar('_NewNumberType', bound=Union[int, float, complex])
+  >>> _NumberType = TypeVar('_NumberType')
+  >>> _NewNumberType = TypeVar('_NewNumberType')
 
-  >>> class Number(Mappable1[_NumberType]):
+  >>> class Number(
+  ...     Kind1['Number', _NumberType],
+  ...     Mappable1[_NumberType],
+  ... ):
   ...     def __init__(self, inner_value: _NumberType):
   ...         self._inner_value = inner_value
   ...
@@ -94,18 +98,18 @@ With our ``Number`` mappable class we can compose easily math functions with it.
 Laws
 ~~~~
 
-To make sure your functor implementation is right, you can apply the
-Functors law on it to test.
+To make sure your mappable implementation is right, you can apply the
+Mappable laws on it to test.
 
 1. **Identity Law:** When we pass the identity function to the map method,
-the functor has to be the same, unaltered.
+the mappable has to be the same, unaltered.
 
 .. code:: python
 
   >>> from returns.functions import identity
 
-  >>> functor = Number(1) # Number[int]
-  >>> assert functor.map(identity) == Number(1)
+  >>> mappable_number: Number[int] = Number(1)
+  >>> assert mappable_number.map(identity) == Number(1)
 
 2. **Associative Law**: Given two functions, `x` and `y`, calling the map
 method with `x` function and after that calling with `y` function must have the
@@ -119,8 +123,8 @@ same result if we compose them together.
   >>> def multiply_by_ten(number: int) -> int:
   ...     return number * 10
 
-  >>> functor = Number(9) # Number[int]
-  >>> assert functor.map(add_one).map(multiply_by_ten) == functor.map(lambda value: multiply_by_ten(add_one(value)))
+  >>> mappable_number: Number[int] = Number(9)
+  >>> assert mappable_number.map(add_one).map(multiply_by_ten) == mappable_number.map(lambda value: multiply_by_ten(add_one(value)))
 
 Naming convention
 -----------------
--- a/docs/pages/interfaces.rst
+++ b/docs/pages/interfaces.rst
@@ -126,6 +126,116 @@ same result if we compose them together.
   >>> mappable_number: Number[int] = Number(9)
   >>> assert mappable_number.map(add_one).map(multiply_by_ten) == mappable_number.map(lambda value: multiply_by_ten(add_one(value)))
 
+Bindable
+--------
+
+Bindable is something that we can bind it with a function, like
+:class:`Maybe <returns.maybe.Maybe>`, so
+:class:`Bindable <returns.interfaces.bindable.BindableN>` interface will help
+us to create our custom bindable.
+
+.. code:: python
+
+  >>> from typing import Any, Callable, TypeVar
+
+  >>> from returns.interfaces.bindable import Bindable1
+  >>> from returns.primitives.hkt import SupportsKind1
+
+  >>> _BagContentType = TypeVar('_BagContentType')
+  >>> _NewBagContentType = TypeVar('_NewBagContentType')
+
+  >>> class Bag(
+  ...     SupportsKind1['Bag', int],
+  ...     Bindable1[_BagContentType],
+  ... ):
+  ...     def __init__(self, inner_value: _BagContentType):
+  ...         self._inner_value = inner_value
+  ...
+  ...     def bind(
+  ...         self,
+  ...         function: Callable[[_BagContentType], 'Bag[_NewBagContentType]']
+  ...     ) -> 'Bag[_NewBagContentType]':
+  ...         return function(self._inner_value)
+  ...
+  ...     def __eq__(self, other: Any) -> bool:
+  ...         if not isinstance(self, type(other)):
+  ...             return False
+  ...         return self._inner_value == other._inner_value
+
+  >>> class Peanuts:
+  ...     def __init__(self, quantity: int) -> None:
+  ...         self.quantity = quantity
+  ...
+  ...     def __eq__(self, other: Any) -> bool:
+  ...         if not isinstance(self, type(other)):
+  ...             return False
+  ...         return self.quantity == other.quantity
+
+  >>> def get_half(peanuts: Peanuts) -> Bag[Peanuts]:
+  ...     return Bag(Peanuts(peanuts.quantity // 2))
+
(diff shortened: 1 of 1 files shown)


docs/pages/interfaces.rst
```diff
@@ -3,8 +3,101 @@
...
+  >>> # We have a shortcut for each amount of arguments to reduce the boilerplate
+  >>> one_type: Mappable1[int]
+  >>> two_types: Mappable2[int, str]
+  >>> three_type: Mappable3[int, str, bool]
+
+.. note::
+
+  Useful links before you start here:
+
+  * `Functors, Applicatives, And Monads In Pictures <http://adit.io/posts/2013-04-17-functors,_applicatives,_and_monads_in_pictures.html>`_
```
sobolevn, line 33: Let's replace "functor", "monad" etc with our own names: mappable, bindable, etc.
See #451

docs/pages/interfaces.rst
```diff
@@ -3,8 +3,101 @@
...
+  * `Understanding Functor and Monad With a Bag of Peanuts <https://medium.com/beingprofessional/understanding-functor-and-monad-with-a-bag-of-peanuts-8fa702b3f69e>`_
+  * `Variance of generic types <https://mypy.readthedocs.io/en/latest/generics.html#variance-of-generic-types>`_
+  * `If you know map, I will teach you monads <https://www.youtube.com/watch?v=cB0vpg9-YMQ>`_
+
+Mappable
+--------
+
+:class:`Mappable <returns.interfaces.mappable.MappableN>` interface help us to
+create our own ``Functor`` like :class:`Result <returns.maybe.Maybe>`.
```
sobolevn, line 42: Let's focus on the interface. `Mappable` is something with `.map` method. That's it.

And then, we can explain why this method is helpful in general.

sobolevn: Btw, 
```
:class:`Result <returns.maybe.Maybe>`
```

Should be: 
```
:class:`Maybe <returns.maybe.Maybe>`
```

docs/pages/interfaces.rst
```diff
@@ -3,8 +3,101 @@
...
+.. code:: python
+
+  >>> from typing import Any, Callable, TypeVar
+
+  >>> from returns.interfaces.mappable import Mappable1
+
+  >>> _FunctorValueType = TypeVar('_FunctorValueType', covariant=True)
+  >>> _NewFunctorValueType = TypeVar('_NewFunctorValueType')
+
+  >>> class MyFunctor(Mappable1[_FunctorValueType]):
```
sobolevn, line 53: Let's give it a unique name, like `Number`

And we will end up illustarting how one can compose `Number`s and math functions.

docs/pages/interfaces.rst
```diff
@@ -3,8 +3,101 @@
...
+  ...             return False
+  ...         return self._inner_value == other._inner_value
+
+Laws
+~~~~
+
+To make sure your functor implementation is right, you can apply the
+Functors law on it to test.
+
+1. **Identity Law:** When we pass the identity function to the map method,
```
sobolevn, line 77: I would suggest to explain **why** do we need this laws in the first place.
Let's show some bad situations when these laws are not respected.

Then, we can show that these laws can save you from yourself. 
Related #457

docs/pages/interfaces.rst
thepabloaguilar: `Functors, Applicatives, And Monads In Pictures` is the article's name, can be a little weird change the name to `Mappable, Applicative, Bindable in Pictures`. The user can be confused hahaha

But in the other scenarios, I agree with you!

sobolevn: Oh, sorry. Yes, articles can be named whatever they originally have in their title.

thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
```diff
@@ -38,36 +38,53 @@ for example, can have one, two or three possible types. See the example below:
 Mappable
 --------
 
-:class:`Mappable <returns.interfaces.mappable.MappableN>` interface help us to
-create our own ``Functor`` like :class:`Result <returns.maybe.Maybe>`.
+Something is considered mappable if we can `map` it using a function, generally
```
sobolevn, line 41: A formatting fix, not:

```
`map`
```

but 

```
``map``
```

docs/pages/interfaces.rst
```diff
@@ -38,36 +38,53 @@ for example, can have one, two or three possible types. See the example below:
...
+`map` is a method that accepts a function. An example in this library is
+:class:`Maybe <returns.maybe.Maybe>`, it has the `map` method.
```
sobolevn, line 43:
```
An example in this library is
:class:`Maybe <returns.maybe.Maybe>`, it has the `map` method:

.. code:: python

   >>> # example body

```

docs/pages/interfaces.rst
```diff
@@ -38,36 +38,53 @@ for example, can have one, two or three possible types. See the example below:
...
 
 .. code:: python
 
   >>> from typing import Any, Callable, TypeVar
 
   >>> from returns.interfaces.mappable import Mappable1
 
-  >>> _FunctorValueType = TypeVar('_FunctorValueType', covariant=True)
-  >>> _NewFunctorValueType = TypeVar('_NewFunctorValueType')
+  >>> _NumberType = TypeVar('_NumberType', covariant=True)
```
sobolevn, line 55: We can remove `covariant=True` here, I guess that it is not important, but simplifies the example.

docs/pages/interfaces.rst
```diff
@@ -38,36 +38,53 @@ for example, can have one, two or three possible types. See the example below:
...
+  >>> _NewNumberType = TypeVar('_NewNumberType')
 
-  >>> class MyFunctor(Mappable1[_FunctorValueType]):
-  ...     _inner_value: _FunctorValueType
+  >>> class Number(Mappable1[_NumberType]):
+  ...     _inner_value: _NumberType
```
sobolevn, line 59: We can skip `_inner_value` annotation here.

docs/pages/interfaces.rst
sobolevn: Let's also bound `_NumberType` to `Union[int, float, complex]`

docs/pages/interfaces.rst
```diff
@@ -38,36 +38,53 @@ for example, can have one, two or three possible types. See the example below:
...
   ...
   ...     def map(  # This method is required by Mappable
   ...         self,
-  ...         function: Callable[[_FunctorValueType], _NewFunctorValueType]
-  ...     ) -> 'MyFunctor[_NewFunctorValueType]':
+  ...         function: Callable[[_NumberType], _NewNumberType]
+  ...     ) -> 'Number[_NewNumberType]':
   ...         new_value = function(self._inner_value)
-  ...         return MyFunctor(new_value)
+  ...         return Number(new_value)
```
sobolevn, line 69: Let's make a single line here, it is way more ideomatic.

docs/pages/interfaces.rst
```diff
@@ -38,36 +38,53 @@ for example, can have one, two or three possible types. See the example below:
...
   ...
   ...     def __eq__(self, other: Any) -> bool:  # Required to check the identity law
   ...         if not isinstance(self, type(other)):
   ...             return False
   ...         return self._inner_value == other._inner_value
 
+With our ``Number`` mappable class we can compose easily math functions on it.
```
sobolevn, line 76: `with it`

docs/pages/interfaces.rst
```diff
@@ -38,36 +38,53 @@ for example, can have one, two or three possible types. See the example below:
...
+
+.. code:: python
+
+  >>> from math import floor, sqrt
+
+  >>> def my_math_function(number: int) -> int:
+  ...     return number + 1
+
+  >>> number: Number[int] = Number(-82)
+  >>> assert number.map(my_math_function).map(abs).map(sqrt).map(floor) == Number(9)
```
sobolevn, line 86: Just two is enough, currently it is hard to follow

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: Do you have some bad situations examples? And a link explain why we need the laws?
Currently I don't have this material!

docs/pages/interfaces.rst
```diff
@@ -38,52 +38,58 @@ for example, can have one, two or three possible types. See the example below:
...
 
 .. code:: python
 
-  >>> from typing import Any, Callable, TypeVar
+  >>> from returns.maybe import Maybe, Some
+
+  >>> def mappable(string: str) -> str:
+  ...     return string + '!'
+
+  >>> maybe_str: Maybe[str] = Maybe.from_value('example')
```
sobolevn, line 54: It is better to use the same constructors on the both sides: `Maybe.from_value` or `Some`.

So:

```python
maybe_str: Maybe[str] = Some('example')
assert maybe_str.map(mappable) == Some('example!')
```

docs/pages/interfaces.rst
```diff
@@ -38,52 +38,58 @@ for example, can have one, two or three possible types. See the example below:
...
+
+.. code:: python
+
+  >>> from typing import Any, Callable, TypeVar, Union
 
   >>> from returns.interfaces.mappable import Mappable1
 
-  >>> _NumberType = TypeVar('_NumberType', covariant=True)
-  >>> _NewNumberType = TypeVar('_NewNumberType')
+  >>> _NumberType = TypeVar('_NumberType', bound=Union[int, float, complex])
```
sobolevn, line 66: Related https://github.com/dry-python/returns/issues/474

sobolevn:
Writting docs is really hard! 😅

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: I've removed the `bound` parameter!

thepabloaguilar:
> Writting docs is really hard! 😅

Yes, it's but can help us to improve our explanation skills, maybe! 😄 😄

docs/pages/interfaces.rst
sobolevn: I actually don't have an experience with broken monads, but let's think of it right now 😅 
I can only think of `Promise` from `js` that sometimes is very hard to use because of it. Like https://buzzdecafe.github.io/2018/04/10/no-promises-are-not-monads

@astynax can you please help us here? Do you have any bad examples in mind?

sobolevn: I also invite @morozzzko to collaborate!

sobolevn: Here are some example of why it is bad: https://mmhaskell.com/monads/laws

Morozzzko: Well, I can clearly see a confusion here. Mainly because I'm confused.

How do we explain **why** an implementation must respect the rule, if the rule is an important part of the definition and the interface? 

I mean, it's just a part of the abstraction. A convention which we use to make it useful. If we want our mappable thing to be useful, it has to follow some rules. Otherwise it just becomes extremely hard to reason about & difficult to work with. 

Let's say we have a mappable which does not respect the identity law. What does it do then? What does it map to? Why does it do that? What's the design benefit? 

I think it's way better not to focus on the **why** and just mention that those lows _make the abstraction work_. 

Perhaps, the docs may ask to send an email if anybody's got a legitimate case when the law must be broken. Would be extremely educational for all parts

sobolevn: We can also give an example where `fmap id != id`, because we can create an implementation where `fmap id` produces a side-effect

thepabloaguilar: I'll add this code below as an example:
```python
class Number(...):
    def map(self, function: Callable[[_Type], _NewType]) -> 'Number[_NewType]':
        return Number(2 * function(self._inner_value))
```

docs/pages/interfaces.rst
```diff
@@ -3,8 +3,234 @@
...
+the mappable has to be the same, unaltered.
+
+.. code:: python
+
+  >>> from returns.functions import identity
+
+  >>> mappable_number: Number[int] = Number(1)
+  >>> assert mappable_number.map(identity) == Number(1)
+
+2. **Associative Law**: Given two functions, `x` and `y`, calling the map
```
sobolevn, line 114:
```
`x`
```

->

```
``x``
```

sobolevn: and the same for every other code block 👍

docs/pages/interfaces.rst
```diff
@@ -3,8 +3,234 @@
...
+.. code:: python
+
+  >>> def add_one(number: int) -> int:
+  ...     return number + 1
+
+  >>> def multiply_by_ten(number: int) -> int:
+  ...     return number * 10
+
+  >>> mappable_number: Number[int] = Number(9)
+  >>> assert mappable_number.map(add_one).map(multiply_by_ten) == mappable_number.map(lambda value: multiply_by_ten(add_one(value)))
```
sobolevn, line 127: What do you think, should we use `compose` here?

docs/pages/interfaces.rst
```diff
@@ -3,8 +3,234 @@
...
+  >>> from typing import Any, Callable, TypeVar
+
+  >>> from returns.interfaces.bindable import Bindable1
+  >>> from returns.primitives.hkt import Kind1
+
+  >>> _BagContentType = TypeVar('_BagContentType')
+  >>> _NewBagContentType = TypeVar('_NewBagContentType')
+
+  >>> class Bag(
+  ...     Kind1['Bag', int],
```
sobolevn, line 149: `SupportsKind1`

docs/pages/interfaces.rst
```diff
@@ -3,8 +3,234 @@
...
+  >>> from typing import Any, Callable, TypeVar
+
+  >>> from returns.interfaces.mappable import Mappable1
+  >>> from returns.primitives.hkt import Kind1
+
+  >>> _NumberType = TypeVar('_NumberType')
+  >>> _NewNumberType = TypeVar('_NewNumberType')
+
+  >>> class Number(
+  ...     Kind1['Number', _NumberType],
```
sobolevn, line 71: `SupportsKind1`

docs/pages/interfaces.rst
```diff
@@ -3,8 +3,234 @@
...
+  ...         return self._inner_value == other._inner_value
+
+  >>> @dataclass
+  ... class Peanuts:
+  ...     quantity: int
+
+  >>> def get_half(peanuts: Peanuts) -> Bag[Peanuts]:
+  ...     return Bag(Peanuts(peanuts.quantity // 2))
+
+  >>> bag_of_peanuts: Bag[Peanuts] = Bag(Peanuts(10))
```
sobolevn, line 173: Awesome example! 👏

docs/pages/interfaces.rst
```diff
@@ -3,8 +3,234 @@
...
+  >>> assert bag.bind(minus_one).bind(half) == bag.bind(lambda value: minus_one(value).bind(half))
+
+What's the difference between ``Mappable`` and ``Bindable``?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+While Mappable you have to pass a pure function, like:
+
+.. code:: python
+
+  >>> def mappable(string: str) -> str:
```
sobolevn, line 223: Let's not name functions `mappable` and `bindable` here. Readers might confuse them with `Mappable` and `Bindable` interfaces. Let's maybe name them `can_be_mapped` and `can_be_bound`?

docs/pages/interfaces.rst
thepabloaguilar: I think so, maybe it'll be more clear to understand the `law`!!

thepabloaguilar:
I've decided to split this PR because is too many information to just one PR! (#517)

sobolevn:
We can work from here! Thanks a lot!

You have built a great foundation for the interfaces docs!

PR #493, opened 2020-07-25
Add a helper to test traces

### LINKED ISSUE #430: Add a helper to test traces
We should add a new method to `pytest` plugin. We need to test traces.

Here's the usecase we are trying to cover. Imagine, there are multiple places where `Failure` can happen with the same result. For example:

```python
def first() -> ResultE[int]:
    return Failure(KeyError('a'))

def second() -> ResultE[int]:
    return Failure(KeyError('a'))
```

And we need to test that the reason why our app failed is `first()` call, not `second()` 
Matching by value is not good enough here, since values are identical. But, we can use traces!

```python
def test_failure(returns):
    assert returns.has_trace(my_function(), 'some trace object with `first()` call')
```

And it will pass for `Failure` created by `first()` and fail for `Failure` created by `second()`

Related to #409 
CC @thepabloaguilar

thepabloaguilar:
The final API I've suggested is like `pytest.raises`:

```python
from returns.result import Result, Success, Failure

def desired_function(arg: str) -> Result[int, str]:
    if arg.isnumeric():
        return Success(int(arg))
    return Failure('"{0}" is not a number'.format(arg))

def test_if_failure_is_created_at_convert_function(returns):
    with returns.has_trace(Failure, desired_function):
        Success('not a number').bind(desired_function)
```

### Why it's better than the initial idea when our plugin will call the function to the user?

It's better because we do not have to call the function to the user!!

---

I've heavily inspired by `pytest` implementation of `raises` but simplifying a little.

* [raises](https://github.com/pytest-dev/pytest/blob/3a060b77e81ebf990159e59cc8f8de26ad998e12/src/_pytest/python_api.py#L568)
* [RaisesContext](https://github.com/pytest-dev/pytest/blob/3a060b77e81ebf990159e59cc8f8de26ad998e12/src/_pytest/python_api.py#L720): This is the `contextmanager`

If the check fails our plugin does not broke the `pytest` flow because we're using `pytest.fail` to communicate the failure:
```text
========================================================================================================== short test summary info ===========================================================================================================
FAILED tests/test_contrib/test_pytest/test_plugin_has_trace.py::test_failed_has_trace_with_one_function_call_in_the_call_stack[Success-Failure] - Failed: No container Success was created
FAILED tests/test_contrib/test_pytest/test_plugin_has_trace.py::test_failed_has_trace_with_one_function_call_in_the_call_stack[IOSuccess-IOFailure] - Failed: No container IOSuccess was created
FAILED tests/test_contrib/test_pytest/test_plugin_has_trace.py::test_failed_has_trace_with_one_function_call_in_the_call_stack[Failure-Success] - Failed: No container Failure was created
FAILED tests/test_contrib/test_pytest/test_plugin_has_trace.py::test_failed_has_trace_with_one_function_call_in_the_call_stack[IOFailure-IOSuccess] - Failed: No container IOFailure was created
======================================================================================================== 4 failed, 12 passed in 1.18s ========================================================================================================
```

Closes #430

### DIFF
--- a/docs/pages/contrib/pytest_plugins.rst
+++ b/docs/pages/contrib/pytest_plugins.rst
@@ -73,6 +73,8 @@ created and looking for the desired function.
 
   >>> from returns.result import Result, Success, Failure
 
+  >>> returns_fixture = getfixture('returns')
+
   >>> def desired_function(arg: str) -> Result[int, str]:
   ...     if arg.isnumeric():
   ...         return Success(int(arg))
@@ -85,6 +87,9 @@ created and looking for the desired function.
   >>> def test_if_success_is_created_at_convert_function(returns):
   ...     with returns.has_trace(Success, desired_function):
   ...         Success('42').bind(desired_function)
+
+  >>> test_if_failure_is_created_at_convert_function(returns_fixture)
+  >>> test_if_success_is_created_at_convert_function(returns_fixture)
 
 Further reading
 ---------------
--- a/docs/pages/contrib/pytest_plugins.rst
+++ b/docs/pages/contrib/pytest_plugins.rst
@@ -60,6 +60,32 @@ This is how it works internally:
   It does not affect production code.
 
 
+has_trace
+~~~~~~~~~
+
+Sometimes we have to know if a container is created correctly in a specific
+point of our flow.
+
+``has_trace`` helps us to check exactly this by identifying when a container is
+created and looking for the desired function.
+
+.. code:: python
+
+  >>> from returns.result import Result, Success, Failure
+
+  >>> def desired_function(arg: str) -> Result[int, str]:
+  ...     if arg.isnumeric():
+  ...         return Success(int(arg))
+  ...     return Failure('"{0}" is not a number'.format(arg))
+
+  >>> def test_if_failure_is_created_at_convert_function(returns):
+  ...     with returns.has_trace(Failure, desired_function):
+  ...         Success('not a number').bind(desired_function)
+
+  >>> def test_if_success_is_created_at_convert_function(returns):
+  ...     with returns.has_trace(Success, desired_function):
+  ...         Success('42').bind(desired_function)
+
 Further reading
 ---------------
 
--- a/returns/contrib/pytest/plugin.py
+++ b/returns/contrib/pytest/plugin.py
@@ -3,10 +3,13 @@ import sys
 from contextlib import contextmanager
 from functools import partial, wraps
 from types import FrameType
-from typing import Any, Callable, Iterator, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, Iterator, TypeVar
 
 import pytest
 from typing_extensions import Final, final
+
+if TYPE_CHECKING:
+    from returns.interfaces.specific.result import ResultBasedN
 
 _ERROR_FIELD: Final = '_error_handled'
 _ERROR_HANDLERS: Final = (
@@ -18,11 +21,9 @@ _ERRORS_COPIERS: Final = (
 )
 
 _FunctionType = TypeVar('_FunctionType', bound=Callable)
-# _ResultBasedType represents `Callable[[Any], ResultBasedN[Any, Any, Any]]`
-# but we cannot import `ResultBasedN` here because we're omitting this file from
-# the coverage report and `coverage.py` will exclude the class' coverage
-# information if we import it here.
-_ResultBasedType = TypeVar('_ResultBasedType', bound=Callable)
+_ResultBasedType = TypeVar(
+    '_ResultBasedType', bound=Callable[..., 'ResultBasedN'],
+)
 
 
 class _DesiredFunctionFound(RuntimeError):
--- a/returns/contrib/pytest/plugin.py
+++ b/returns/contrib/pytest/plugin.py
@@ -21,8 +21,8 @@ _ERRORS_COPIERS: Final = (
 )
 
 _FunctionType = TypeVar('_FunctionType', bound=Callable)
-_ResultBasedType = TypeVar(
-    '_ResultBasedType', bound=Callable[..., 'ResultBasedN'],
+_ResultCallableType = TypeVar(
+    '_ResultCallableType', bound=Callable[..., 'ResultBasedN'],
 )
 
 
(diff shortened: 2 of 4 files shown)


docs/pages/contrib/pytest_plugins.rst
```diff
@@ -60,6 +60,32 @@ This is how it works internally:
...
+.. code:: python
+
+  >>> from returns.result import Result, Success, Failure
+
+  >>> def desired_function(arg: str) -> Result[int, str]:
+  ...     if arg.isnumeric():
+  ...         return Success(int(arg))
+  ...     return Failure('"{0}" is not a number'.format(arg))
+
+  >>> def test_if_failure_is_created_at_convert_function(returns):
```
sobolevn, line 81: These test functions right now are not called. Am I right?
Or does `pytest` execute them?

returns/contrib/pytest/plugin.py
```diff
@@ -13,6 +17,17 @@
     'alt',
 )
 
+_FunctionType = TypeVar('_FunctionType', bound=Callable)
+# _ResultBasedType represents `Callable[[Any], ResultBasedN[Any, Any, Any]]`
+# but we cannot import `ResultBasedN` here because we're omitting this file from
+# the coverage report and `coverage.py` will exclude the class' coverage
+# information if we import it here.
+_ResultBasedType = TypeVar('_ResultBasedType', bound=Callable)
```
sobolevn, line 25: Can we use 

```python
if TYPE_CHECKING:
   _ResultCallableType = TypeVar('_ResultBasedType', bound=Callable[..., ResultBasedN])
```

?

sobolevn: And then use it like `'_ResultCallableType'`

docs/pages/contrib/pytest_plugins.rst
thepabloaguilar: You're right, `pytest` does not execute them!

returns/contrib/pytest/plugin.py
thepabloaguilar: The problem is the import statment, but following your idea we can do this:
```python
if TYPE_CHECKING:
    from returns.interfaces.specific.result import ResultBasedN

_ResultCallableType = TypeVar(
    '_ResultCallableType', bound=Callable[..., 'ResultBasedN'],
)
```

docs/pages/contrib/pytest_plugins.rst
sobolevn: Then, let's call them manually!

returns/contrib/pytest/plugin.py
sobolevn: Yes, that's exactly what I've meant 🙂

docs/pages/contrib/pytest_plugins.rst
thepabloaguilar: How can I do this?? Because we're not able to "instantiate" the fixtures manually like:
```python
from returns.contrib.pytest.plugin import returns as returns_plugin, _patch_containers

returns_fixture = returns_plugin(_patch_containers())
```
Output:
```text
Fixture "_patch_containers" called directly. Fixtures are not meant to be called directly,
but are created automatically when test functions request them as parameters.
See https://docs.pytest.org/en/latest/fixture.html for more information about fixtures, and
https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly about how to update your code.
```

thepabloaguilar:
> One more question: what about other types:
> 
> * `ReaderResult` uses `Result`, so we should probably test it too
> * `ReaderIOResult`, the same
> * `FutureResult` uses `async def (a) -> Result[b, c]` as a type, and returns `IOResult`. Should we also test it?
> * The same question about `ReaderFutureResult`

I'm not sure if we have to test them because they are not the `{IO}Result` itself, they uses `{IO}Result` internally or return it. In the end everyone will handle a `ResultBasedN` or `IOResultBasedN`.

returns/contrib/pytest/plugin.py
```diff
@@ -18,11 +21,9 @@
 )
 
 _FunctionType = TypeVar('_FunctionType', bound=Callable)
-# _ResultBasedType represents `Callable[[Any], ResultBasedN[Any, Any, Any]]`
-# but we cannot import `ResultBasedN` here because we're omitting this file from
-# the coverage report and `coverage.py` will exclude the class' coverage
-# information if we import it here.
-_ResultBasedType = TypeVar('_ResultBasedType', bound=Callable)
+_ResultCallableType = TypeVar(
+    '_ResultCallableType', bound=Callable[..., 'ResultBasedN'],
```
sobolevn, line 25: Just a quick question: should we add `.trace` property to `ResultBasedN`?

thepabloaguilar: I think so, but our feature `collect_traces` will not work with custom Results because we monkey patch `_Failure` directly!

sobolevn: Ok, then it should not be a part of the interface 👍

docs/pages/contrib/pytest_plugins.rst
sobolevn: See https://docs.pytest.org/en/stable/doctest.html#using-fixtures

returns/contrib/pytest/plugin.py
thepabloaguilar: I'll make some tests latter, I think we can achieve the trace for custom Results by monkey patching `ResultBasedN` directly!!
But the trace will be available in all types like `_Success`, it can be a problem or not since `collect_traces` should be used only for development purposes.

sobolevn: I don't think that we need this. Because this is an implementation detail. I am not sure that other people might need it.

docs/pages/contrib/pytest_plugins.rst
thepabloaguilar: Done!

sobolevn:
Thanks a lot! 👍

sobolevn:
@thepabloaguilar oups, we forgot to add a changelog entry!

thepabloaguilar:
> @thepabloaguilar oups, we forgot to add a changelog entry!

I'm going to add

Issue #488, opened 2020-07-21
Possible mistake in doc: map used instead of bind


nurumaik:
I believe these lines https://github.com/dry-python/returns/blame/master/README.md#L109
```python
user: Optional[User]
# Type hint here is optional, it only helps the reader here:
discount_program: Maybe['DiscountProgram'] = Maybe.from_value(
    user,
).map(  # This won't be called if `user is None`
    lambda real_user: real_user.get_balance(),
).map(  # This won't be called if `real_user.get_balance()` returns None
    lambda balance: balance.credit_amount(),
).map(  # And so on!
    lambda credit: choose_discount(credit) if credit > 0 else None,
)
```

Should use bind instead of map, because every lambda has type x -> Maybe[y], not x -> y

sobolevn:
Yes! You are right, that's what happens when you don't actually test your examples!

Are you willing to fix that? 😄

nurumaik:
Can do this on weekends

sobolevn:
Great, thanks!

Issue #314, opened 2020-04-09
Unexpected behavior of `Some(None)`


homeworkprod:
I expect the `Some` "constructor" (actually just a function) to only return `Some`s.

However, given `None`:
```py
>>> from returns.maybe import Some
>>> Some(None)
<returns.maybe._Nothing object at 0x7f345dcaa1c8>
```

For reference, another language that has null pointers, Java, raises an exception on `Optional.of(null)`.

In general, I expect the explicitly requested creation of a `Some` to fail if no actual value to wrap in it is supplied.

If one explicitly wants to let `Maybe` decide whether to create a `Some` or `None` given a thing, there still is `Maybe.new` (to which `Some` just delegates).

astynax:
Should stay `Some(None)` definitely.

homeworkprod:
@astynax, thanks for looking into it.

However, I'm quite confused about your verdict, to be honest.

As I understand the [option type](https://en.wikipedia.org/wiki/Option_type), the two states it represents are either nothing or a present value.

`Some(None)` however would be a container that wraps nothing.

Pattern matching is a popular choice to unwrap a maybe/an option type, like this (example from Wikipedia page mentioned earlier):
```haskell
-- This function uses pattern matching to deconstruct Maybes
compute :: Maybe Int -> String
compute Nothing  = "No value"
compute (Just x) = "The value is: " ++ show x
```

The equivalent of `Some(None)` can not even be represented in Haskell (correct me if I'm wrong). `Just Nothing` is possible, but the *returns* equivalent is `Some(Nothing)`, but *not* `Some(None)`.

The whole point of a "something" object is to guarantee that it contains something. `None` is not something.

Also, the docstring of *returns*' [`_Some` class](https://github.com/dry-python/returns/blob/master/returns/maybe.py#L244) states:
`Represents a calculation which has succeeded and contains the value.`
I think it's clear that `None` in Python world is not a value.

I might be overlooking something here, but `Some(None)` seems totally unacceptable to me.

Is it possible you meant `Some(Nothing)` instead? Because that is something I think would make sense. Then again, I would only allow it to be constructed via `Some(Nothing)`, but not as a fallback for `Some(None)` – which, to me, still is an invalid value to pass to `Some`, and worth a `TypeError` or `ValueError` to be raised.

astynax:
[...]

PR #494, opened 2020-07-26
Use bind instead of map in example

nurumaik:
Read the docs, and saw no need of updating and explaining bind right there, so I'm modifying only the code part

https://github.com/dry-python/returns/issues/488

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,11 +111,11 @@ user: Optional[User]
 # Type hint here is optional, it only helps the reader here:
 discount_program: Maybe['DiscountProgram'] = Maybe.from_value(
     user,
-).map(  # This won't be called if `user is None`
+).bind(  # This won't be called if `user is None`
     lambda real_user: real_user.get_balance(),
-).map(  # This won't be called if `real_user.get_balance()` returns None
+).bind(  # This won't be called if `real_user.get_balance()` returns None
     lambda balance: balance.credit_amount(),
-).map(  # And so on!
+).bind(  # And so on!
     lambda credit: choose_discount(credit) if credit > 0 else None,
 )
 ```


nurumaik:
Idk how to link issue though

nurumaik:
So your library is doing smth weird indeed, just as I thought first

map on Optional[T] is working like bind on Maybe[T]. My brain expected that if we map None to Nothing and x to Some x, then map method should work like this:

```python
Some(1).map(lambda x: return x * 2) == Some(2)
Some(1).map(lambda x: return None) == Some(None)
Some(1).bind(lambda x: return None) == Nothing
```

Guess all that weirdness comes from python "type" system, sorry for the confusion :)

sobolevn:
Yes, 
```bash
» python -c 'from returns.maybe import Some; print(Some(1).map(lambda _: None))'
<Nothing>
```

This is intentional. Because `Some(None)` is a really strange thing to have in Python. 
See #314

sobolevn:
Hey, @nurumaik! 👋 

Thanks again for submitting this fix! And for raising the discussion.
I hope that it is now clear, what is going on in this example. I will 100% add a note about `Some(None)` to it and then close #488 

Feel free to ask any questions you have! 👍

nurumaik:
> Hey, @nurumaik! 👋
> 
> Thanks again for submitting this fix! And for raising the discussion.
> I hope that it is now clear, what is going on in this example. I will 100% add a note about `Some(None)` to it and then close #488
> 
> Feel free to ask any questions you have! 👍

Sorry, it was too hard to change documentation back, so I changed the code to conform to my changes in docs instead

https://github.com/dry-python/returns/pull/510

Issue #491, opened 2020-07-25
Create `Unit` type alias for `None`


sobolevn:
Sometimes it might be helpful to make a differce between `-> None` and `-> Unit`.

We need to investigate how it is possible.

Examples:

```python
def some() -> Unit:
    print('Just prints')

class Some():
    def __init__(self) -> None: ...
```

sobolevn:
Or, we can use `()` as a unit type https://github.com/python/mypy/issues/4211

sobolevn:
This will be very hard to read for newcovers. Let's leave it as good-old `None`.

By the way, `()` won't work as a unit value. You would have to explicitly return it everywhere. Not a great idea.

PR #510, opened 2020-07-28
WIP: fixing compositions in Maybe

### LINKED ISSUE #488: Possible mistake in doc: map used instead of bind

### LINKED ISSUE #501: Add MaybeBasedN interface
Base types:
- `Mappable`
- `Bindable`
- `Applicative`

Declarations:
- `or_else_call`

That's it!

nurumaik:
I don't expect it to be merged (bc it breaks compatibility), just sharing my thoughts on how proper Maybe should work

1. Some(None) is as normal as [[]] and [None]. Fact that it doesn't make sense don't mean that users shouldn't be allowed to create it
2. There is one vital property of map that's violated in this library. Map should ALWAYS preserve the structure. Some(x).map(any_function) should always be equal to Some(y). Like list of 5 elements mapped to some function should be also list of 5 elements
3. All practical problems with it are solved by allowing bind (because it's natural for bind to do that) to promote non-maybe values to Optional[] and then to Maybe[] using usual conversion

Closes #488 
Closes #501


nurumaik:
http://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Functor.html#t:Functor

A type f is a Functor if it provides a function fmap which, given any types a and b lets you apply any function from (a -> b) to turn an f a into an f b, **preserving the structure of f**

Obligatory reference to the language of gods, so you won't think that I came up with that idea myself

sobolevn:
@nurumaik thanks a lot for this! 

Here's some history behind this decision:
1. `Maybe` was originally working like this for a long time
2. Then we got a question from a user: #314 
3. Then I have tried to implement the samething you do here: https://github.com/dry-python/returns/issues/314
4. I have reverted this commit, because of: https://github.com/dry-python/returns/issues/314#issuecomment-629775417

The problem here is that `None` does not translate properly to Haskell terms. We have:
- `None` as a unit value `()`
- `None` as a `null` / `nil` value 

And I am ok to break the `map` law here, because otherwise, `Maybe` would be too hard to deal with in Python.
That's something the language dictates us to do.

What's your opinion on that?

nurumaik:
Disclaimer: I'm not a library user(but very intrested in what it does) and don't even use python as a main language, so this is more of a discussion than request

For map and bind currently works like
map(f: X->Y) -> Maybe[Y]
bind(f: X->Maybe[Y]) -> Maybe[Y]
map(f: X->Optional[Y]) -> Maybe[Y]
bind(f: X->Optional[Y]) -> Error

while I suggest just changing last 2 behaviors to
map(f: X->Optional[Y]) -> Maybe[Optional[Y]]
bind(f: X->Optional[Y]) -> Maybe[Y]

Now you can use bind where you would use map and have same result, and map/bind behavior on Optionals is the same as with Maybes

For dealing with external optional values, and not ending up with Maybe[Optional[X]] my proposal is to leave Maybe.from_value as is, but allow creating Some directly from None. So you use Maybe.from_value for external libraries which may return None, and Some with your own None-free code

> And I don't want to have Maybe[Optional[int]] this is really ugly.
Well, if your map has None values, then result of this function would be indeed Maybe[Optional[int]], because it's a result of a querying dict by key, where Nothing is "not found", Some(None) is "None found" and Some(1) is "1 found"
If you are sure that there would be no Nones in the library, then it's safe to just declare return type as Maybe[int]
If you want to get your Nones as Nothing, you use Maybe.from_value instead of Some

Maybe I'm missing smth about python type system here, and "just declare return type as Maybe[int]" wouldn't typecheck, then the proper solution would be to add another strictly converting method, like Some.from_value or Maybe.from_value_strict, which will throw exception if value is None

Breaking Functor laws in current Maybe has very deep implications. For example, composition doesn't work anymore:
```
f = lambda x: None
g = lambda x: 2
fg = lambda x: g(f(x))
Maybe.from_value(1).map(f).map(g) == Nothing
Maybe.from_value(1).map(fg) == Some(2)
```

And I believe that it is pretty real to stumble on this while refactoring some complex code. Experienced "functional" (in a sense of haskell and monads) programmer would expect that it's safe to replace .map(f).map(g) with .map(fg) while it actually isn't 

>The problem here is that None does not translate properly to Haskell terms
Here's an example of how scala solves similar problem with nulls https://scastie.scala-lang.org/1ZKBOqrwSdWbdiiazhyNHQ. I mainly drew inspiration from scala implementation of Option here

So, I don't see it impossible to keep library sane while also preserving important algebraic laws

sobolevn:
I really like your `bind_optional` suggestion. 

> And I believe that it is pretty real to stumble on this while refactoring some complex code. Experienced "functional" (in a sense of haskell and monads) programmer would expect that it's safe to replace .map(f).map(g) with .map(fg) while it actually isn't

Yes, I agree on this one two.

> strict Some.from_value

I also like this idea as well. 

Would you want to work further on this?
I think that you have found a way to combine correctness and usability in one solution.

nurumaik:
Yes, I'll just need some time to study python type system deeper

sobolevn:
Might be related #491

sobolevn:
And also #501

Morozzzko:
Hello, Ruby user here. We have our own [monads at dry-rb](https://dry-rb.org/gems/dry-monads/), so here's a little bit of my own experience with monads in a dynamically typed language.

Now, for the further context. Let's map something from returns / Python to dry-monads / Ruby for a better understanding

| returns | dry-monads |
| --- | -- |
| `None` | `nil` |
| `Some` | `Some` |
| `Nothing` | `None` |

Alright. Now I'm going to provide a little snippet with REPL which illustrates how this specific case works for us:

```
irb(main):001:0> require 'dry/monads/all'
=> true
irb(main):002:0> require 'dry/monads/version'
=> true
irb(main):003:0> Dry::Monads::VERSION
=> "1.3.5"
irb(main):004:0> Dry::Monads.Some(nil)
        7: from /Users/moroz/.asdf/installs/ruby/2.6.5/bin/irb:23:in `<main>'
        6: from /Users/moroz/.asdf/installs/ruby/2.6.5/bin/irb:23:in `load'
        5: from /Users/moroz/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        4: from (irb):4
        3: from /Users/moroz/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/dry-monads-1.3.5/lib/dry/monads/maybe.rb:264:in `Some'
        2: from /Users/moroz/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/dry-monads-1.3.5/lib/dry/monads/maybe.rb:264:in `new'
        1: from /Users/moroz/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/dry-monads-1.3.5/lib/dry/monads/maybe.rb:108:in `initialize' as
ArgumentError (nil cannot be some)
irb(main):005:0> Dry::Monads.Some(1).fmap { nil }
=> None
```

What you can see here: You can't get `Some` of `nil`, yet you can get `nil` via `fmap`, which breaks the laws.

However, I'd argue that the proposed behavior should be implemented in a different manner. See the approach in dry-monads: https://github.com/dry-rb/dry-monads/pull/105

They just added another method, which _for now_ aliases to fmap and does the coercion. Later versions _will not allow_ blocks to return `nil`.

I firmly believe `returns` should follow the same path. Here's why.

Just like Ruby, Python is a dynamically typed language which has its own special value: `None`. In some cases, it serves the same purpose as `Nothing`, in others it does not.

However, `None` is still an artifact of a _different_ type system. When we're using Monads, we agree to use some specific approaches and reduce our use of alternative approaches.

Which means, in an ideal situation, we should never write code which combines `Maybe` and `None` because it's harder to reason about. What exactly is the meaning of `None` here? Is it `Nothing`? Is it `Unit`? By using this library we've agreed to add a little explicitness to our code, and we should probably follow it.

Yet, the library should allow its users to follow the correct path and enable proper composition. While it should _probably_ not be the default, it's good to have a method which works like `map`, but doesn't coerce the value.

sobolevn:
Ok, without this feature we would not be able to build `NonEmptySequence` container.
Because `non_empty_sequence.get(0) == Nothing` can happen for two reasons:
1. sequence is empty
2. sequence has `None` as the first element

We need `Some(None)` to be possible to represent. I really like what @Morozzzko described. I guess it is the right for both ruby and python.

Morozzzko:
@sobolevn sorry, I'm _really_ confused about two topics.

You're saying that you need this feature to implement a NonEmptySequence because `get(0)` may return Nothing in two cases, and one of them is _the sequence is empty_. I can't see how this is valid

Also, you're referring to my message and claiming that `Some(None)` should be possible to represent. Yet, Ruby's implementation is going to explicitly prohibit it in `fmap` instead of enabling it

nurumaik:
> Hello, Ruby user here. We have our own [monads at dry-rb](https://dry-rb.org/gems/dry-monads/), so here's a little bit of my own experience with monads in a dynamically typed language.
> 
> Now, for the further context. Let's map something from returns / Python to dry-monads / Ruby for a better understanding
> returns 	dry-monads
> `None` 	`nil`
> `Some` 	`Some`
> `Nothing` 	`None`
> 
> Alright. Now I'm going to provide a little snippet with REPL which illustrates how this specific case works for us:
> 
> ```
> irb(main):001:0> require 'dry/monads/all'
> => true
> irb(main):002:0> require 'dry/monads/version'
> => true
> irb(main):003:0> Dry::Monads::VERSION
> => "1.3.5"
> irb(main):004:0> Dry::Monads.Some(nil)
> Traceback (most recent call last):
>         7: from /Users/moroz/.asdf/installs/ruby/2.6.5/bin/irb:23:in `<main>'
>         6: from /Users/moroz/.asdf/installs/ruby/2.6.5/bin/irb:23:in `load'
>         5: from /Users/moroz/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
>         4: from (irb):4
>         3: from /Users/moroz/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/dry-monads-1.3.5/lib/dry/monads/maybe.rb:264:in `Some'
>         2: from /Users/moroz/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/dry-monads-1.3.5/lib/dry/monads/maybe.rb:264:in `new'
>         1: from /Users/moroz/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/dry-monads-1.3.5/lib/dry/monads/maybe.rb:108:in `initialize' as
> ArgumentError (nil cannot be some)
> irb(main):005:0> Dry::Monads.Some(1).fmap { nil }
> => None
> ```
> 
> What you can see here: You can't get `Some` of `nil`, yet you can get `nil` via `fmap`, which breaks the laws.
> 
> However, I'd argue that the proposed behavior should be implemented in a different manner. See the approach in dry-monads: [dry-rb/dry-monads#105](https://github.com/dry-rb/dry-monads/pull/105)
> 
> They just added another method, which _for now_ aliases to fmap and does the coercion. Later versions _will not allow_ blocks to return `nil`.
> 
> I firmly believe `returns` should follow the same path. Here's why.
> 
> Just like Ruby, Python is a dynamically typed language which has its own special value: `None`. In some cases, it serves the same purpose as `Nothing`, in others it does not.
> 
> However, `None` is still an artifact of a _different_ type system. When we're using Monads, we agree to use some specific approaches and reduce our use of alternative approaches.
> 
> Which means, in an ideal situation, we should never write code which combines `Maybe` and `None` because it's harder to reason about. What exactly is the meaning of `None` here? Is it `Nothing`? Is it `Unit`? By using this library we've agreed to add a little explicitness to our code, and we should probably follow it.
> 
> Yet, the library should allow its users to follow the correct path and enable proper composition. While it should _probably_ not be the default, it's good to have a method which works like `map`, but doesn't coerce the value.

Ok, imagine you have some third-party code producing some value and some other third-party code consuming that value. Your program doesn't look at this value, only passing it around using some logic (involving wrapping it in Maybe at some point). Then, libraries change api slightly, including None in the possible set of values. Of course your program breaks because you don't allow Some(None). Because Option disallowing creation of Some(None) is leaky abstration and leaky abstractions are called bad thing for a reason

Morozzzko:
> Then, libraries change api slightly, including None in the possible set of values. Of course your program breaks because you don't allow Some(None). Because Option disallowing creation of Some(None) is leaky abstration and leaky abstractions are called bad thing for a reason

That's not really a problem of the abstraction. Making _any_ value nullable is a potentially breaking action in most modern programming languages. Statically typed languages which have a built-in concept of _optional types_ won't even compile the code.

Since extending a type by adding `None` is never a safe operation, it's much better to have an option to learn about it as fast as possible. Exception on `None` is a good solution, as it would clearly indicate that the program is trying to do something you haven't designed it for.

However, if we decide to just allow `Some(None)` without extra explicitness, it may lead to corrupted data. Which is

a) more difficult to trace than an exception
b) may stay undetected for who-knows-how-long

This is honestly a lot worse than not being fully category-theory compatible. 

Let's take a look at another languages, where we express optional values with `null` or a similar value. Java introduced [Optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html), which is its own version of the same concept. Let's see the behavior here.

Here's the description of map:

> If a value is present, and the value matches the given predicate, return an Optional describing the value, otherwise return an empty Optional.

So, let's see how this works. I made this snippet:

```java
import java.util.Optional;

class Main {
  public static void main(String[] args) {
    Optional<Integer> value = Optional.of(5);
    Optional<Integer> mappedValue = value.map(_x -> null);
    System.out.println(mappedValue);
  }
}
```

The returned value? `Optional.empty`.

This is a demonstration how a different implementation of a same concept behaves this way. I'm not saying it's good, but I can totally say that **this behavior is predictable** for people who are familiar with it. 

Here's the thing: once someone decides to use `Optional` or `Maybe` in their code, they should agree to reduce their use of `None` / `null` and _switch to optional_. Those two are _conflicting_.

sobolevn:
So, here's the API I propose: https://gist.github.com/sobolevn/a860b7e00e58298b1fcdfbb21ae4f7a2

Key points:
- `_Some(None)` is possible, but `Some(None)` is not possible
- `Maybe[Optional[int]]` is possible
- `map()` never changes container type
- new `opt()` introduced, that behaves like current `map()`, it can change a container type

This design in my opinion is:
1. Semantically correct, all laws will pass on it
2. Pythonic, since we minimize the `Some(None)` confusion
3. Useful for validation, because we can now see the difference between `{'a': None}.get('a')` and `{'a': None}.get('b')`

Feedback is welcome!

thepabloaguilar:
It's better to rename `opt` to `optional`/`map_optional`, seems more clear!

---

`mypy` can't prevent the users to pass `None` if we have a `TypeVar`, so

```python
from typing import TYPE_CHECKING, TypeVar, Optional

if not TYPE_CHECKING:
    reveal_type = print

MyType = TypeVar('MyType')

def function_one(arg: MyType) -> MyType:
  return arg

def function_two(arg: Optional[MyType]) -> Optional[MyType]:
  return arg

def function_three(arg: int) -> int:
  return arg

function_one(None)
function_two(None)
function_three(None)
```

Output:
```text
example.py:32: error: Argument 1 to "function_three" has incompatible type "None"; expected "int"
Found 1 error in 1 file (checked 1 source file)
```

---

I don't like much the idea of `.from_value(...)` behaviors different from `Some(...)` because all of `returns` containers works like:
* Result.from_value(...) == Success(...)
* Result.from_failure(...) == Failure(...)

sobolevn:
@thepabloaguilar thanks a lot for your reasonable feedback!

> It's better to rename opt to optional/map_optional, seems more clear!

It thougt that `opt` is good enough for two reasons:
1. consistency in length with `map` and `alt`
2. `opt` as a verb has a similar meaning https://www.macmillandictionary.com/dictionary/british/opt

I don't have a strong opinion here, so let's try it out.

> mypy can't prevent the users to pass None if we have a TypeVar, so

mypy plugin can! 🙂 

> I don't like much the idea of .from_value(...) behaves different from Some(...) because all of returns containers works like

Great point! But, there are several important differences here:
1. `ResultLikeN` has both `.from_value` and `.from_failure, while `Maybe` has just `.from_value`
2. Both `Success` and `Failure` are functions, while `Nothing` is a singleton

This seems like a completely different use-case to me, doesn't it?

thepabloaguilar:
Just one more point here, what about if we explicitly add two methods:
* `.from_value(...)`
* `.from_optional_value(...)`

sobolevn:
@thepabloaguilar can you please explain a bit more? What is the difference between these two methods?

thepabloaguilar:
> @thepabloaguilar can you please explain a bit more? What is the difference between these two methods?

Desconsider my last comment, please! I think it's not a great idea hahaha

sobolevn:
Now I see why `.from_optional` is required. Otherwise, `.from_value` breaks a bunch of laws:

```
============================== short test summary info ===============================
FAILED ex.py::test_maybe_containern_left_identity_law - AssertionError: <Nothing> =...
FAILED ex.py::test_maybe_applicativen_homomorphism_law - AssertionError: <Some: Fal...
FAILED ex.py::test_maybe_applicativen_interchange_law - AssertionError: <Nothing> =...
============================ 3 failed, 9 passed in 19.09s ============================

```

thepabloaguilar:
My idea when I proposed two methods was to maintain `.from_value` like others without any treatments and `.from_optional` treating if the value is None!

sobolevn:
TODO:
- [ ] type tests for `Maybe` interfaces
- [ ] `bind_optional` method and pointfree
- [ ] type tests for `bind_optional`
- [ ] `Maybe` laws

sobolevn:
@nurumaik your review is very welcome!

sobolevn:
I will leave this open due Saturday. And then merge to continue my work on other features.

thepabloaguilar (approved):
LGTM

PR #527, opened 2020-07-31
Fixes `Result.alt(...)` documentation

### LINKED ISSUE #516: Documentation issue (`Result.alt(...)` method)
Wrong return type in the `Result` container documentation of `.alt(...)`:
```python
>>> from returns.result import Result, Failure, Success

>>> def altable(arg: str) -> Result[str, str]:
...      return arg + 'b'

>>> assert Success('a').alt(altable) == Success('a')
>>> assert Failure('a').alt(altable) == Failure('ab')
```

The signature of `altable` should be:
```python
>>> def altable(arg: str) -> str:
...      return arg + 'b'
```

thepabloaguilar:
closes #516

### DIFF
--- a/returns/result.py
+++ b/returns/result.py
@@ -207,9 +207,9 @@ class Result(
 
         .. code:: python
 
-          >>> from returns.result import Result, Failure, Success
+          >>> from returns.result import Failure, Success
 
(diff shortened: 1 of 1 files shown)


sobolevn:
Thanks!

PR #529, opened 2020-08-02
Adds `cond` method

### LINKED ISSUE #437: Consider adding `.cond` to `Result` container
[Either](https://typelevel.org/cats/datatypes/eithert.html) provides a way to avoid the boilerplate below:
```python
def func():
    if something:
        return Success(...)
    return Failure(...)
```

It provides the `.cond` method, with `Result` container will be something like this:
```python
def func():
    return Result.cond(
        my_condition,   # The boolean value
        success_value,  # If boolean is `True` the return will be Success(success_value)
        failure_value,  # If boolean is `False` the return will be Failure(failure_value)
    )
```

thepabloaguilar:
Working code example:
```python
>>> from returns.methods import cond
>>> from returns.result import Failure, Result, Success

>>> def is_numeric(string: str) -> Result[str, str]:
...     return cond(
...         Result,
...         string.isnumeric(),
...         'It is a number',
...         'It is not a number',
...     )

>>> assert is_numeric('42') == Success('It is a number')
>>> assert is_numeric('non numeric') == Failure('It is not a number')
```

Closes #437

### DIFF
--- a/returns/methods/cond.py
+++ b/returns/methods/cond.py
@@ -1,0 +1,45 @@
+from typing import Type, TypeVar
+
+from returns.interfaces.specific.result import ResultLikeN
+from returns.primitives.hkt import KindN, kinded
+
+_ValueType = TypeVar('_ValueType')
+_ErrorType = TypeVar('_ErrorType')
+_ThirdType = TypeVar('_ThirdType')
+
+_ResultKind = TypeVar('_ResultKind', bound=ResultLikeN)
+
+
+def internal_cond(
+    container_type: Type[_ResultKind],
+    is_success: bool,
+    success_value: _ValueType,
+    error_value: _ErrorType,
+) -> KindN[_ResultKind, _ValueType, _ErrorType, _ThirdType]:
+    """
+    Help us to reduce the boilerplate when choosing paths with ``ResultLikeN``.
+
+    .. code:: python
+
+      >>> from returns.methods import cond
+      >>> from returns.result import Failure, Result, Success
+
+      >>> def is_numeric(string: str) -> Result[str, str]:
+      ...     return cond(
+      ...         Result,
+      ...         string.isnumeric(),
+      ...         'It is a number',
+      ...         'It is not a number',
+      ...     )
+
+      >>> assert is_numeric('42') == Success('It is a number')
+      >>> assert is_numeric('non numeric') == Failure('It is not a number')
+
+    """
+    if is_success:
+        return container_type.from_value(success_value)
+    return container_type.from_failure(error_value)
+
+
+#: Kinded version of :func:`~internal_cond`, use it to infer real return type.
+cond = kinded(internal_cond)
--- a/returns/methods/cond.py
+++ b/returns/methods/cond.py
@@ -1,11 +1,10 @@
 from typing import Type, TypeVar
 
 from returns.interfaces.specific.result import ResultLikeN
-from returns.primitives.hkt import KindN, kinded
+from returns.primitives.hkt import Kind2, kinded
 
 _ValueType = TypeVar('_ValueType')
 _ErrorType = TypeVar('_ErrorType')
-_ThirdType = TypeVar('_ThirdType')
 
 _ResultKind = TypeVar('_ResultKind', bound=ResultLikeN)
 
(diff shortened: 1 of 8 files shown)


typesafety/test_methods/test_cond.yml
```diff
@@ -0,0 +1,7 @@
+- case: cond_1
+  disable_cache: true
+  main: |
+    from returns.methods import cond
+    from returns.result import Result
+
+    reveal_type(cond(Result, True, 42, '42'))  # N: Revealed type is 'returns.result.Result*[builtins.int*, builtins.str*]'
```
sobolevn, line 7: Let's add `FutureResult`,`IOResult`, `ReaderIOResult`, `ReaderFutureResult`, and custom datatype test-cases.

thepabloaguilar:
> 1. What do you think about point-free function for this method?

Can be handy, so we need to add `.cond(...)` method to our interface right??

```python
class ResultLikeN(...):
    @classmethod
    def cond(
        cls, boolean: bool, success_value: _ST, error_value: _ET,
    ) -> 'ResultLikeN[_ST, _ET, _TT]':
        if boolean:
            return cls.from_value(success_value)
        return cls.from_error(error_value)

def cond(
    boolean: bool,
    success_value: _ST,
    error_value: _ET,
) -> Callable[[_ResultKind], KindN[_ResultKind, _ST, _ET, _TT]]:
    return lambda container: container.cond(boolean, success_value, error_value)
```

> 2. Can you please add a changelog entry?

Yes, I can!

sobolevn:
@thepabloaguilar no, I was asking about https://github.com/dry-python/returns/tree/master/returns/pointfree

thepabloaguilar:
Oh, I didn't see that changes. I remember every pointfree function were implement around container methods!!
Now these functions are implemented around the functions from the `methods` package.

thepabloaguilar:
So, the API should be:
```python
cond(Result, success, failure)(is_success)
```

thepabloaguilar:
Does `point free cond` might be included in other PR?

sobolevn:
@thepabloaguilar yes, please! 👍

typesafety/test_methods/test_cond.yml
```diff
@@ -0,0 +1,62 @@
...
+
+    reveal_type(cond(ReaderIOResult, True, 1.0, False))  # N: Revealed type is 'returns.context.requires_context_ioresult.RequiresContextIOResult*[builtins.float*, builtins.bool*, <nothing>]'
+
+- case: cond_reader_future_result
+  disable_cache: true
+  main: |
+    from returns.methods import cond
+    from returns.context import ReaderFutureResult
+
+    reveal_type(cond(ReaderFutureResult, True, 1, 1.0))  # N: Revealed type is 'returns.context.requires_context_future_result.RequiresContextFutureResult*[builtins.int*, builtins.float*, <nothing>]'
```
sobolevn, line 39: Should be `returns.context.requires_context_future_result.RequiresContextFutureResult*[builtins.int*, builtins.float*, Any]`

returns/methods/cond.py
```diff
@@ -0,0 +1,45 @@
...
+
+_ResultKind = TypeVar('_ResultKind', bound=ResultLikeN)
+
+
+def internal_cond(
+    container_type: Type[_ResultKind],
+    is_success: bool,
+    success_value: _ValueType,
+    error_value: _ErrorType,
+) -> KindN[_ResultKind, _ValueType, _ErrorType, _ThirdType]:
```
sobolevn, line 18: Should be `Kind2`, because we don't work with the third type at all

sobolevn: That's why it is always `never`

typesafety/test_methods/test_cond.yml
thepabloaguilar: I think we have a `type` problem here or not!! It just work if I assign the result to a variable and typing it properly:
```yml
- case: cond_reader_ioresult
  disable_cache: true
  main: |
    from typing import Any

    from returns.methods import cond
    from returns.context import ReaderIOResult

    container: ReaderIOResult[float, bool, Any] = cond(ReaderIOResult, True, 1.0, False)
    reveal_type(container)  # N: Revealed type is 'returns.context.requires_context_ioresult.RequiresContextIOResult[builtins.float, builtins.bool, Any]'
```

thepabloaguilar: Sorry, it's working without the above solution!! It was `mypy` cache kidding me hahaha
I just deleted the cache an everything worked properly!!

PR #538, opened 2020-08-07
Adds documentation about the difference between `ResultLikeN` and `ResultBasedN`

### LINKED ISSUE #536: `is_successful` should accepts `ResultLikeN`
`is_successful` accepts only `Unwrappable` types, but this is not semantic at all.
Now, the big difference between `ResultLikeN` and `ResultBasedN` is that we can unwrap the second one but we have to be able to use `is_successful` with `ResultLikeN` because it represents computation paths (success, failure) too.

And I think this is not a great function name to use with `Maybe`, `is_empty` and `is_nothing` are better names and can be a separated function! Essentially `Maybe` tells us if we have something or not.

thepabloaguilar:
Creates the content of `What is the difference between ResultLikeN and ResultBasedN?` section, in `interfaces.rst`

Closes #536

### DIFF
--- a/docs/pages/interfaces.rst
+++ b/docs/pages/interfaces.rst
@@ -256,13 +256,13 @@ What is the difference between ResultLikeN and ResultBasedN?
 ``ResultLikeN`` is just an intention of having a result
 (e.g. :class:`FutureResult <returns.future.FutureResult>`),
 it's not the result yet. While ``ResultBasedN`` is a concret result
-(e.g. :class:`Result <returns.result.Result>`),
+(e.g. :class:`IOResult <returns.io.IOResult>`),
 it's has the desired result value.
 
 Because of this difference between them is why we can't unwrap a ``ResultLikeN``
 container, it does not have the real result yet.
 
-See the example below using ``FutureResult`` to get a ``Result``:
+See the example below using ``FutureResult`` to get a ``IOResult``:
 
 .. code:: python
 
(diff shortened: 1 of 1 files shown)


docs/pages/interfaces.rst
```diff
@@ -253,6 +253,42 @@ Why some interfaces do not have type alias for 1 type argument?
 What is the difference between ResultLikeN and ResultBasedN?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+``ResultLikeN`` is just an intention of having a result
+(e.g. :class:`FutureResult <returns.future.FutureResult>`),
+it's not the result yet. While ``ResultBasedN`` is a concret result
+(e.g. :class:`Result <returns.result.Result>`),
```
sobolevn, line 259: Let's use `IOResult` here as an example

docs/pages/interfaces.rst
```diff
@@ -253,6 +253,42 @@ Why some interfaces do not have type alias for 1 type argument?
...
+
+  >>> async def coro(arg: int) -> Result[int, str]:
+  ...     return Success(arg + 1)
+
+  >>> # `result_like` does not have the result we want (Result[int, str])
+  >>> # it's just the intention of having one, we have to await it to get the real result
+  >>> result_like: FutureResult[int, str] = FutureResult(coro(1))
+  >>> assert isinstance(result_like, ResultLikeN)
+  >>> # `anyio.run(...)` will await our coroutine and give the real result to us
+  >>> result: Result[int, str] = anyio.run(result_like.awaitable)
```
sobolevn, line 284: It will return `IOResult`, not `Result`

sobolevn:
Thank you!

PR #548, opened 2020-08-13
Adds `swap` as a method

### LINKED ISSUE #513: `swap` should be a method, not a converter
Following the same logic: we have `.swap()` method in containers, just like any other `.bind` or similar methods.

thepabloaguilar:
Closes #513


returns/converters.py
```diff
@@ -16,53 +16,6 @@
 _ResultLikeKind = TypeVar('_ResultLikeKind', bound=ResultLikeN)
```
sobolevn, line 16: Looks like this is unused.

thepabloaguilar: Oh, yes. Done!

PR #561, opened 2020-08-24
Moves aliases to interfaces

### LINKED ISSUE #551: Move aliases to interfaces
Creating a separate `aliases/` subpackage was a mistake.
We need to join them with basic interfaces.

thepabloaguilar:
Btw, I've removed the `abstractproperty` decorator because it's deprecated since Python 3.3, [Reference](https://docs.python.org/3/library/abc.html#abc.abstractproperty)

The "new style":
```python
class Example(ABC):
    @property
    @abstractmethod
    def my_abstract_property(self):
        ...
```

Closes #551


returns/interfaces/specific/reader.py
```diff
@@ -117,7 +117,8 @@ def __call__(self, deps: _ThirdType) -> Any:
         Or any other type.
         """
 
-    @abstractproperty
+    @property
```
sobolevn, line 120: Great! I was unaware of this!

sobolevn (approved):
Great work!

PR #567, opened 2020-08-26
Adds tooltips for docs

### LINKED ISSUE #537: Consider adding popups for docs
Like hypothesis has: https://hypothesis.readthedocs.io/en/latest/data.html
Снимок экрана 2020-08-06 в 11 27 50

thepabloaguilar:
We don't need to change our references manually, it's automatically since `hoverxref_auto_ref` option is `True`!!

**Disclaimer**: Locally it's do not works because we need to set `hoverxref_api_host`, `hoverxref_project`and `hoverxref_version` explicitly. But in the RTD build environment will work normally because the plugin uses the env variables from RTD. [Configuration reference](https://sphinx-hoverxref.readthedocs.io/en/latest/configuration.html).
We can set the variables like `hoverxref_* = os.getenv('READTHEDOCS_*) or 'default_value'`, but I think is not good and we don't need this in our CI or locally it's a little overhead because `hoverxref` uses the RTD API to get the information about the references.
I've tested locally setting the variables properly, everything worked fine.

Btw, I found some wrong references like [here](https://github.com/dry-python/returns/pull/567/files#diff-1533ff41e93cd154fdb05ba1c5229037R72) maybe a plugin to verify these references will be great.

Closes #537


sobolevn:
> Btw, I found some wrong references like here maybe a plugin to verify these references will be great.

Yes! Can you please open a new issue for it? 
I don't know how to do it.

sobolevn (approved):

sobolevn:
Oh, sorry. I have merged something else and now you have to do the rebase 😞

PR #569, opened 2020-08-28
Adds `map vs bind` to `FAQ` section

### LINKED ISSUE #480: Need explanation for the usecase of map_ and .map
Hi, first I just want to thank you for this awesome library :)

I'm wondering about the map_ function and its unsafe usage. I might be doing this wrong, if so please correct me!

So if i'm making a flow i could do something like this:

```python
# read file and load json to dict
flow(
    'filename.json',
    read_function,  # returns ResultE[str]
    map_(json.loads),
    rescue(handle_error)
)
```
Lets say reading the file works nicely, but actually the json is malformed which makes map_(json.loads) throw an exception that is _not_ caught by the flow. It will also not end up in any `alt` nor `rescue` parts of the flow. 

Another example:
```python
Success('some_param').map(int).fix(len)  # fix will not be called here
```
changing from `map(int)` to `bind(safe(int))` does work, but then we're back to using `bind`.

So my question is what is the usecase for .map and map_ if they cannot be called safely?

Now this library is all about Railway Oriented Programming and I think it would make sense to catch the exceptions in the examples and return a Failure, which then can propogate down to `alt` or `rescue` part of the flow. How this should be done idk, but basically the same as the @safe annotation does would be nice.

I also think this makes sense since it would unify what we are working on when writing flows. that the map function returns a Modal will make it similar to bind. and nested Modals won't be something you'd have to think about that much.

I might be doing things wrong but basically all functions or callable objects i write when using this library returns a Modal. So when composing those with flow or pipe I will be using bind. Only when calling builtins I will use map. So I believe that builtins or other functions from other libraries that does not return a modal should be _forced to do it_ in a flow. And the way to force it should be with `map_`/`.map`

thepabloaguilar:
I don't know it's a good example, I've used your example about `json.loads`!!

Closes #480

### DIFF
--- a/docs/pages/result.rst
+++ b/docs/pages/result.rst
@@ -181,6 +181,35 @@ Like so:
 
 So, that's a way to go, if you need this composition.
 
+map vs bind
+~~~~~~~~~~~
+
+We use the ``map`` method when we're working with pure functions, a function
+is pure if it doesn't produce any side-effect (e.g. Exceptions). On the other
+hand, we use the ``bind`` method if a function returns a ``Result`` instance
+which translates its potential side-effect into a raw value.
+See the example below:
+
+.. code:: python
+
+  >>> import json
+  >>> from typing import Dict
+
+  >>> from returns.result import Failure, Result, Success, safe
+
+  >>> # `cast_to_bool` doesn't produce any side-effect
+  >>> def cast_to_bool(arg: int) -> bool:
+  ...     return bool(arg)
+
+  >>> # `parse_json` can produce Exceptions, so we use the `safe` decorator
+  >>> # to prevent any kind of exceptions
+  >>> @safe
+  ... def parse_json(arg: str) -> Dict[str, str]:
+  ...     return json.loads(arg)
+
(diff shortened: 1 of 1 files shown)


docs/pages/result.rst
```diff
@@ -181,6 +181,34 @@ Like so:
 
 So, that's a way to go, if you need this composition.
 
+map vs bind
+~~~~~~~~~~~
+
+We use the ``map`` method when we're working with pure functions, a function
+is pure if it doesn't produce any side-effect (e.g. Exceptions). On the other
+hand, we use the``bind`` method if a function can produce side-effects.
```
sobolevn, line 189: > we use the``bind`` method if a function returns a `Result` instance which translates its potential side effect into a raw value

sobolevn (changes requested):
Let's fix one wording!

Thanks a lot 👍

docs/pages/result.rst
thepabloaguilar: Done!

sobolevn (approved):

sobolevn:
Thanks a lot!

PR #581, opened 2020-09-05
Removes all methods

sobolevn:
It was very easy to remove. 

We can keep them (extra code to maintain, extra tests to run, extra complexity).
Or we can remove them with possibly no costs.


sobolevn:
Ok, no `xdist` is good enough. But it is much simplier setup.

PR #614, opened 2020-09-18
Active `fail_on_warning` RTD

### LINKED ISSUE #568: Active `fail_on_warning` RTD
Now we're ignoring the `sphinx` warnings and some of these warnings is about "broken" references so we need to configure our `ReadTheDocs` build to fail on warnings.
The default value is __false__, [reference](https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx-fail-on-warning)!!

Another point here is that we got a lot of warnings like this:
```text
WARNING: Cannot resolve forward reference in type annotations of "returns.interfaces.specific.result.ResultLikeN.bind_result": name 'Result' is not defined
```
It's because our configuration is set to disable the `TYPE_CHECKING` flag as we can see [here](https://github.com/dry-python/returns/blob/master/docs/conf.py#L109)

refs #547

thepabloaguilar:
I've fixed all documentation warnings and created `.readthedocs.yml` to active the `fail_on_warning` flag!!
But we have a problem here, to fix the warnings about `ForwardReference` I've added the imports `from __future__ import annotations`, unfortunately the hypothesis' tests are failing. I think it's related to this issue: https://github.com/HypothesisWorks/hypothesis/issues/2565

## Checklist

## Related issues

Closes #568

### DIFF
--- a/returns/context/requires_context.py
+++ b/returns/context/requires_context.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -87,7 +89,7 @@ class RequiresContext(
     """
 
     #: This field has an extra 'RequiresContext' just because `mypy` needs it.
-    _inner_value: Callable[['RequiresContext', _EnvType], _ReturnType]
+    _inner_value: Callable[[RequiresContext, _EnvType], _ReturnType]
 
     #: A convenient placeholder to call methods created by `.from_value()`:
     empty: ClassVar[NoDeps] = object()
@@ -136,7 +138,7 @@ class RequiresContext(
 
     def map(  # noqa: WPS125
         self, function: Callable[[_ReturnType], _NewReturnType],
-    ) -> 'RequiresContext[_NewReturnType, _EnvType]':
+    ) -> RequiresContext[_NewReturnType, _EnvType]:
         """
         Allows to compose functions inside the wrapped container.
 
@@ -164,7 +166,7 @@ class RequiresContext(
             Callable[[_ReturnType], _NewReturnType],
             _EnvType,
         ],
-    ) -> 'RequiresContext[_NewReturnType, _EnvType]':
+    ) -> RequiresContext[_NewReturnType, _EnvType]:
         """
         Calls a wrapped function in a container on this container.
 
@@ -186,7 +188,7 @@ class RequiresContext(
             [_ReturnType],
             Kind2['RequiresContext', _NewReturnType, _EnvType],
         ],
-    ) -> 'RequiresContext[_NewReturnType, _EnvType]':
+    ) -> RequiresContext[_NewReturnType, _EnvType]:
         """
         Composes a container with a function returning another container.
 
@@ -221,7 +223,7 @@ class RequiresContext(
     def modify_env(
         self,
         function: Callable[[_NewEnvType], _EnvType],
-    ) -> 'RequiresContext[_ReturnType, _NewEnvType]':
+    ) -> RequiresContext[_ReturnType, _NewEnvType]:
         """
         Allows to modify the environment type.
 
@@ -238,7 +240,7 @@ class RequiresContext(
         return RequiresContext(lambda deps: self(function(deps)))
 
     @classmethod
-    def ask(cls) -> 'RequiresContext[_EnvType, _EnvType]':
+    def ask(cls) -> RequiresContext[_EnvType, _EnvType]:
         """
         Get current context to use the dependencies.
 
@@ -316,13 +318,13 @@ class RequiresContext(
         See also:
             https://dev.to/gcanti/getting-started-with-fp-ts-reader-1ie5
 
-        """
+        """  # noqa: F811
         return RequiresContext(identity)
 
     @classmethod
     def from_value(
         cls, inner_value: _FirstType,
-    ) -> 'RequiresContext[_FirstType, NoDeps]':
+    ) -> RequiresContext[_FirstType, NoDeps]:
         """
         Used to return some specific value from the container.
 
@@ -343,8 +345,8 @@ class RequiresContext(
 
     @classmethod
     def from_context(
-        cls, inner_value: 'RequiresContext[_NewReturnType, _NewEnvType]',
-    ) -> 'RequiresContext[_NewReturnType, _NewEnvType]':
+        cls, inner_value: RequiresContext[_NewReturnType, _NewEnvType],
+    ) -> RequiresContext[_NewReturnType, _NewEnvType]:
         """
         Used to create new containers from existing ones.
 
@@ -363,10 +365,10 @@ class RequiresContext(
     def from_iterable(
         cls,
         inner_value: Iterable[
-            Kind2['RequiresContext', _NewReturnType, _NewEnvType],
+            Kind2[RequiresContext, _NewReturnType, _NewEnvType],
         ],
         strategy: Type[BaseIterableStrategyN] = FailFast,
-    ) -> 'RequiresContext[Sequence[_NewReturnType], _NewEnvType]':
+    ) -> RequiresContext[Sequence[_NewReturnType], _NewEnvType]:
         """
         Transforms an iterable of ``RequiresContext`` containers.
 
@@ -388,7 +390,7 @@ class RequiresContext(
     def from_requires_context_result(
         cls,
         inner_value: 'RequiresContextResult[_ValueType, _ErrorType, _EnvType]',
-    ) -> 'RequiresContext[Result[_ValueType, _ErrorType], _EnvType]':
+    ) -> RequiresContext[Result[_ValueType, _ErrorType], _EnvType]:
         """
         Typecasts ``RequiresContextResult`` to ``RequiresContext`` instance.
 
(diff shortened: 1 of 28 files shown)


sobolevn:
😭

docs/pages/pointfree.rst
```diff
@@ -285,8 +285,6 @@ API Reference
 
 .. autofunction:: returns.pointfree.bind_context_result
 
-.. autofunction:: returns.pointfree.bind_context_ioresult
```
sobolevn, line 288: Why deleting it?

returns/context/requires_context.py
```diff
@@ -316,7 +318,7 @@ def ask(cls) -> 'RequiresContext[_EnvType, _EnvType]':
         See also:
             https://dev.to/gcanti/getting-started-with-fp-ts-reader-1ie5
 
-        """
+        """  # noqa: F811
```
sobolevn, line 321: Why do we need this `noqa`?

returns/io.py
```diff
@@ -604,7 +606,7 @@ def failure(self) -> IO[_ErrorType]:
             ...
           returns.primitives.exceptions.UnwrapFailedError
 
-        """
+        """  # noqa: RST399
```
sobolevn, line 609: Why do we need this `noqa`?

setup.cfg
```diff
@@ -171,3 +171,4 @@ show_traceback = True
 ignore-path = docs/_build
 max-line-length = 80
 sphinx = True
+ignore-path-errors=docs/pages/railway.rst;D000
```
sobolevn, line 174: Why do we ignore this file?

sobolevn (commented):
Thanks a lot for your work! 

I don't know how to fix Hypothesis types right now.
Because the easiest way is to contribute to Hypothesis the fix we need.
But, the fix we need is non-trivial. Because it should work for all generics, not just `Type`, etc.
And I don't have any solution to offer right now. 
So, if you have any ideas how to solve the type resolutions, please, share them! 👍

returns/interfaces/specific/future.py
```diff
@@ -6,6 +6,8 @@
 :class:`returns.interfaces.specific.future_result.FutureResultBasedN` type.
 """
 
+from __future__ import annotations
```
sobolevn, line 9: I see that you don't change `'Future[_UpdatedType]'` to `Future[_UpdatedType]`
Maybe, that's the problem?

docs/pages/pointfree.rst
thepabloaguilar: Oh, it's because when I removed this line the `bind_context_ioresult` was not there (introduced by #590).
I have rebased my branch yesterday, I'll return this line!!

returns/context/requires_context.py
thepabloaguilar: This is kind weird, after the addition of `annotations` import I got the following error:
```text
./returns/context/requires_context.py

  277:15   F811  redefinition of unused 'Deps' from line 255
  """
        Get current context to use the dependencies.

        It is a common scenario when you need to use the environment.
        For example, you want to do some context-related computation,
        but you don't have the context instance at your disposal.
        That's where ``.ask()`` becomes useful!

        .. code:: python

          >>> from typing_extensions import TypedDict
          >>> class Deps(TypedDict):
          ...     message: str

          >>> def first(lg: bool) -> RequiresContext[int, Deps]:
          ...     # `deps` has `Deps` type here:
          ...     return RequiresContext(
          ...         lambda deps: deps['message'] if lg else 'error',
          ...     )

          >>> def second(text: str) -> RequiresContext[int, int]:
          ...     return first(len(text) > 3)

          >>> assert second('abc')({'message': 'ok'}) == 'error'
          >>> assert second('abcd')({'message': 'ok'}) == 'ok'

        And now imagine that you have to change this ``3`` limit.
        And you want to be able to set it via environment as well.
        Ok, let's fix it with the power of ``RequiresContext.ask()``!

        .. code:: python

          >>> from typing_extensions import TypedDict
          >>> class Deps(TypedDict):
          ...     message: str
          ...     limit: int   # note this new field!

          >>> def new_first(lg: bool) -> RequiresContext[int, Deps]:
          ...     # `deps` has `Deps` type here:
          ...     return RequiresContext(
          ...         lambda deps: deps['message'] if lg else 'err',
          ...     )

          >>> def new_second(text: str) -> RequiresContext[int, int]:
          ...     return RequiresContext[int, Deps].ask().bind(
          ...         lambda deps: new_first(len(text) > deps.get('limit', 3)),
          ...     )

          >>> assert new_second('abc')({'message': 'ok', 'limit': 2}) == 'ok'
          >>> assert new_second('abcd')({'message': 'ok'}) == 'ok'
          >>> assert new_second('abcd')({'message': 'ok', 'limit': 5}) == 'err'

        That's how ``ask`` works.

        This class contains methods that require
        to explicitly set type annotations. Why?
        Because it is impossible to figure out the type without them.
        So, here's how you should use them:

        .. code:: python

            RequiresContext[int, Dict[str, str]].ask()

        Otherwise, your ``.ask()`` method
        will return ``RequiresContext[<nothing>, <nothing>]``,
        which is unusable:

        .. code:: python

            env = RequiresContext.ask()
            env(some_deps)

        And ``mypy`` will warn you: ``error: Need type annotation for '...'``

        See also:
            https://dev.to/gcanti/getting-started-with-fp-ts-reader-1ie5

        """
```

returns/io.py
thepabloaguilar: It's because we need add the `:force:` directive, without this directive `pygments` throw a warning!!
`flake8` was warning because that.

```python
class Immutable(object):
    """
    Helper type for objects that should be immutable.

    When applied, each instance becomes immutable.
    Nothing can be added or deleted from it.

    .. code:: pycon
      :force:  # HERE

      >>> from returns.primitives.types import Immutable
      >>> class MyModel(Immutable):
      ...     ...

      >>> model = MyModel()
      >>> model.prop = 1
         ...
      returns.primitives.exceptions.ImmutableStateError

    See :class:`returns.primitives.container.BaseContainer` for examples.

    """
```

flake8 output:
```text
  579:1    RST399 Error in "code" directive:
  """
        Get value from successful container or raise exception for failed one.

        .. code:: pycon
          :force:

          >>> from returns.io import IO, IOFailure, IOSuccess
          >>> assert IOSuccess(1).unwrap() == IO(1)

          >>> IOFailure(1).unwrap()
            ...
          returns.primitives.exceptions.UnwrapFailedError

        """
  ^
```

setup.cfg
thepabloaguilar: `invalid RST format - D000`

Same root cause of the previous comment, `:force:` directive!!

```rst
.. code:: pycon
  :force:

  >>> assert Failure(1).failure() == 1
  >>> Success(1).failure()
    ...
  returns.primitives.exceptions.UnwrapFailedError
```

doc8 output:
```text
docs/pages/railway.rst:146: D000 Error in "code" directive:
unknown option: "force".
docs/pages/railway.rst:163: D000 Error in "code" directive:
unknown option: "force".
```

returns/interfaces/specific/future.py
thepabloaguilar: I'm happy to say, you're right 🥇

thepabloaguilar: I'll adjust this for all files containing `annotations` import

thepabloaguilar: Ok, I can't do this for all files!!
I got some errors about forward references:
```text
WARNING: Cannot resolve forward reference in type annotations of ...
```

thepabloaguilar: I think I have found the "perfect" balance

thepabloaguilar: I found what I did wrong, my bad. I was removing the apostrophes from types restricted to type checking

thepabloaguilar: Btw, I can't remove the apostrophes from `'Future[_UpdatedType]'` because `Future` is restricted to type checking in this file

returns/context/requires_context.py
sobolevn: Looks like `flake8` issue

returns/io.py
sobolevn: Can you please report it in `flake8-rst-docstrings`? This is not correct.

setup.cfg
sobolevn: The same, this issue should be reported.

sobolevn (approved):

returns/io.py
thepabloaguilar: I'll open the issue, but we'll can remove the `:force:` after the `pygments 2.7.2` asap!
https://github.com/pygments/pygments/issues/1548

thepabloaguilar: https://github.com/peterjc/flake8-rst-docstrings/issues/28

setup.cfg
thepabloaguilar: Seems the issue was already reported: https://github.com/PyCQA/doc8/issues/30

returns/context/requires_context.py
thepabloaguilar: https://github.com/PyCQA/pyflakes/issues/578

PR #624, opened 2020-09-24
Creates `failable` interfaces

### LINKED ISSUE #623: We don't have an interface to represent `Failable` types
Currently we treat `ResultLikeN` as a basic failable:

```python
class ResultLikeN(
    ContainerN[_FirstType, _SecondType, _ThirdType],
    bimappable.BiMappableN[_FirstType, _SecondType, _ThirdType],
    rescuable.RescuableN[_FirstType, _SecondType, _ThirdType],
    Lawful['ResultLikeN[_FirstType, _SecondType, _ThirdType]'],
): ...
```

But, that's not correct. And even our own code supports this statement by having `Maybe`:

```python
class MaybeLikeN(
    ContainerN[_FirstType, _SecondType, _ThirdType],
    rescuable.RescuableN[_FirstType, _SecondType, _ThirdType],
    Lawful['MaybeLikeN[_FirstType, _SecondType, _ThirdType]'],
): ...
```

They are almost identical (except for `.alt` part, but more on this later). So, we need a basic interface to represent both of them.

I can see several things here:
- [ ] We need to create a new `interfaces.failable` module, we would define two types there
- [ ] First type would be `Failable` that extends both `ContainerN` and `Rescueable`. This type is supposed to have only a single failed value
- [ ] Second type would be `DiverseFailable` that extends both `Failable` and `BiMappable`. The thing here is that some types can only fail with one value: `None`, `0`, `False`, `NaN`, you name it. But, other types can fail with literally any value: like `Result` does. So, some types just don't need `.alt` and `.from_failure` methods
- [ ] We need to move `swap` and `.from_failure` from `ResultLikeN` to `DiverseFailable` type
- [ ] Type test all the things!
- [ ] Change docs where we already cover this

thepabloaguilar:
# Creates `failable` interfaces

Instead of have just one `LawSpec` I've made two:
* `_FailableLawSpec` for `FailableN` interface
* `_DiverseFailableLawSpec` for `DiverseFailableN` interface

Please see if the tests are enough, I made based on `ResultLikeN` tests!

I don't find where I have to modify the docs, I've add the API reference, is this?

## Checklist

- [X] I have added my changes to the `CHANGELOG.md`: Already has the entry (Adds new public interfaces: see `returns.interfaces`)

### Feature related

- [X] We need to create a new `interfaces.failable` module, we would define two types there
- [X] First type would be `Failable` that extends both `ContainerN` and `Rescueable`. This type is supposed to have only a single failed value
- [X] Second type would be `DiverseFailable` that extends both `Failable` and `BiMappable`. The thing here is that some types can only fail with one value: `None`, `0`, `False`, `NaN`, you name it. But, other types can fail with literally any value: like `Result` does. So, some types just don't need `.alt` and `.from_failure` methods
- [X] We need to move `swap` and `.from_failure` from `ResultLikeN` to `DiverseFailable` type
- [X] Type test all the things!
- [X] Change docs where we already cover this

## Related issues

Closes #623

### DIFF
--- a/docs/pages/interfaces.rst
+++ b/docs/pages/interfaces.rst
@@ -452,6 +452,7 @@ Here's a full overview of all our interfaces:
    returns.interfaces.bimappable
    returns.interfaces.unwrappable
    returns.interfaces.container
+   returns.interfaces.failable
    returns.interfaces.specific.maybe
    returns.interfaces.specific.result
    returns.interfaces.specific.io
@@ -563,6 +564,16 @@ Container
    :strict:
 
 .. automodule:: returns.interfaces.container
+  :members:
+  :private-members:
+
+Failable
+~~~~~~~~
+
+.. autoclasstree:: returns.interfaces.failable
+   :strict:
+
+.. automodule:: returns.interfaces.failable
   :members:
   :private-members:
 
--- a/returns/interfaces/failable.py
+++ b/returns/interfaces/failable.py
@@ -1,0 +1,261 @@
+from abc import abstractmethod
+from typing import Callable, ClassVar, NoReturn, Sequence, Type, TypeVar
+
+from typing_extensions import final
+
+from returns.interfaces.bimappable import BiMappableN
+from returns.interfaces.container import ContainerN
+from returns.interfaces.rescuable import RescuableN
+from returns.primitives.asserts import assert_equal
+from returns.primitives.hkt import KindN
+from returns.primitives.laws import (
+    Law,
+    Law2,
+    Law3,
+    Lawful,
+    LawSpecDef,
+    law_definition,
+)
+
+_FirstType = TypeVar('_FirstType')
+_SecondType = TypeVar('_SecondType')
+_ThirdType = TypeVar('_ThirdType')
+_UpdatedType = TypeVar('_UpdatedType')
+
+_SingleFailableType = TypeVar('_SingleFailableType', bound='SingleFailableN')
+_DiverseFailableType = TypeVar('_DiverseFailableType', bound='DiverseFailableN')
+
+# Used in laws:
+_NewFirstType = TypeVar('_NewFirstType')
+_NewSecondType = TypeVar('_NewSecondType')
+
+
+@final
+class _FailableLawSpec(LawSpecDef):
+    """
+    Failable laws.
+
+    We need to be sure that ``.rescue`` won't rescue success types.
+    """
+
+    @law_definition
+    def rescue_short_circuit_law(
+        raw_value: _FirstType,
+        container: 'FailableN[_FirstType, _SecondType, _ThirdType]',
+        function: Callable[
+            [_SecondType],
+            KindN['FailableN', _FirstType, _NewFirstType, _ThirdType],
+        ],
+    ) -> None:
+        """Ensures that you cannot rescue a success."""
+        assert_equal(
+            container.from_value(raw_value),
+            container.from_value(raw_value).rescue(function),
+        )
+
+
+class FailableN(
+    ContainerN[_FirstType, _SecondType, _ThirdType],
+    RescuableN[_FirstType, _SecondType, _ThirdType],
+    Lawful['FailableN[_FirstType, _SecondType, _ThirdType]'],
+):
+    """
+    Base type for types that can fail.
+
+    It is a raw type and should not be used directly.
+    Use ``SingleFailableN`` and ``DiverseFailableN`` instead.
+    """
+
+    _laws: ClassVar[Sequence[Law]] = (
+        Law3(_FailableLawSpec.rescue_short_circuit_law),
+    )
+
+
+#: Type alias for kinds with two type arguments.
+Failable2 = FailableN[_FirstType, _SecondType, NoReturn]
+
+#: Type alias for kinds with three type arguments.
+Failable3 = FailableN[_FirstType, _SecondType, _ThirdType]
+
+
+@final
+class _SingleFailableLawSpec(LawSpecDef):
+    """
+    Single Failable laws.
+
+    We need to be sure that ``.map`` and ``.bind``
+    works correctly for ``empty`` property.
+    """
+
+    @law_definition
+    def map_short_circuit_law(
+        container: 'SingleFailableN[_FirstType, _SecondType, _ThirdType]',
+        function: Callable[[_FirstType], _NewFirstType],
+    ) -> None:
+        """Ensures that you cannot map from the `empty` property."""
+        assert_equal(
+            container.empty,
+            container.empty.map(function),
+        )
+
+    @law_definition
+    def bind_short_circuit_law(
+        container: 'SingleFailableN[_FirstType, _SecondType, _ThirdType]',
+        function: Callable[
+            [_FirstType],
+            KindN['SingleFailableN', _NewFirstType, _SecondType, _ThirdType],
+        ],
+    ) -> None:
+        """Ensures that you cannot bind from the `empty` property."""
+        assert_equal(
+            container.empty,
+            container.empty.bind(function),
+        )
+
+    @law_definition
+    def apply_short_circuit_law(
+        container: 'SingleFailableN[_FirstType, _SecondType, _ThirdType]',
+        function: Callable[[_FirstType], _NewFirstType],
+    ) -> None:
+        """Ensures that you cannot apply from the `empty` property."""
+        wrapped_function = container.from_value(function)
+        assert_equal(
+            container.empty,
+            container.empty.apply(wrapped_function),
+        )
+
+
+class SingleFailableN(
+    FailableN[_FirstType, _SecondType, _ThirdType],
+):
+    """
+    Base type for types that have just only one failed value.
+
+    Like ``Maybe`` types where the only failed value is ``Nothing``.
+    """
+
+    _laws: ClassVar[Sequence[Law]] = (
+        Law2(_SingleFailableLawSpec.map_short_circuit_law),
+        Law2(_SingleFailableLawSpec.bind_short_circuit_law),
+        Law2(_SingleFailableLawSpec.apply_short_circuit_law),
+    )
+
+    @property
+    @abstractmethod
+    def empty(
+        self: _SingleFailableType,
+    ) -> 'SingleFailableN[_FirstType, _SecondType, _ThirdType]':
+        """This property represents the failed value."""
+
+
+#: Type alias for kinds with two types arguments.
+SingleFailable2 = SingleFailableN[_FirstType, _SecondType, NoReturn]
+
+#: Type alias for kinds with three type arguments.
+SingleFailable3 = SingleFailableN[_FirstType, _SecondType, _ThirdType]
+
+
+@final
+class _DiverseFailableLawSpec(LawSpecDef):
+    """
+    Diverse Failable laws.
+
+    We need to be sure that ``.map``, ``.bind``, ``.apply`` and ``.alt``
+    works correctly for both success and failure types.
+    """
+
+    @law_definition
+    def map_short_circuit_law(
+        raw_value: _SecondType,
+        container: 'DiverseFailableN[_FirstType, _SecondType, _ThirdType]',
+        function: Callable[[_FirstType], _NewFirstType],
+    ) -> None:
+        """Ensures that you cannot map a failure."""
+        assert_equal(
+            container.from_failure(raw_value),
+            container.from_failure(raw_value).map(function),
+        )
+
+    @law_definition
+    def bind_short_circuit_law(
+        raw_value: _SecondType,
+        container: 'DiverseFailableN[_FirstType, _SecondType, _ThirdType]',
+        function: Callable[
+            [_FirstType],
+            KindN['DiverseFailableN', _NewFirstType, _SecondType, _ThirdType],
+        ],
+    ) -> None:
+        """
+        Ensures that you cannot bind a failure.
+
+        See: https://wiki.haskell.org/Typeclassopedia#MonadFail
+        """
+        assert_equal(
+            container.from_failure(raw_value),
+            container.from_failure(raw_value).bind(function),
+        )
+
+    @law_definition
+    def apply_short_circuit_law(
+        raw_value: _SecondType,
+        container: 'DiverseFailableN[_FirstType, _SecondType, _ThirdType]',
+        function: Callable[[_FirstType], _NewFirstType],
+    ) -> None:
+        """Ensures that you cannot apply a failure."""
+        wrapped_function = container.from_value(function)
+        assert_equal(
+            container.from_failure(raw_value),
+            container.from_failure(raw_value).apply(wrapped_function),
+        )
+
+    @law_definition
+    def alt_short_circuit_law(
+        raw_value: _SecondType,
+        container: 'DiverseFailableN[_FirstType, _SecondType, _ThirdType]',
+        function: Callable[[_SecondType], _NewFirstType],
+    ) -> None:
+        """Ensures that you cannot alt a success."""
+        assert_equal(
+            container.from_value(raw_value),
+            container.from_value(raw_value).alt(function),
+        )
+
+
+class DiverseFailableN(
+    FailableN[_FirstType, _SecondType, _ThirdType],
+    BiMappableN[_FirstType, _SecondType, _ThirdType],
+    Lawful['DiverseFailableN[_FirstType, _SecondType, _ThirdType]'],
+):
+    """
+    Base type for types that have any failed value.
+
+    Like ``Result`` types.
+    """
+
+    _laws: ClassVar[Sequence[Law]] = (
+        Law3(_DiverseFailableLawSpec.map_short_circuit_law),
+        Law3(_DiverseFailableLawSpec.bind_short_circuit_law),
+        Law3(_DiverseFailableLawSpec.apply_short_circuit_law),
+        Law3(_DiverseFailableLawSpec.alt_short_circuit_law),
+    )
+
+    @abstractmethod
+    def swap(
+        self: _DiverseFailableType,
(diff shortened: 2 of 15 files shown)


docs/pages/interfaces.rst
```diff
@@ -564,6 +564,15 @@ IOResult specific
   :members:
   :private-members:
 
+Failable specifc
```
sobolevn, line 567: I think that `Failable` is not specific.

We can interfaces `specific` when they do work with some our existing types.
`Failable` does not work with any of them.

returns/interfaces/specific/maybe.py
```diff
@@ -12,8 +12,8 @@
 
 from typing_extensions import final
 
-from returns.interfaces import equable, rescuable, unwrappable
-from returns.interfaces.container import ContainerN
+from returns.interfaces import equable, unwrappable
```
sobolevn, line 15: `from returns.interfaces import equable, unwrappable, failable`

returns/interfaces/specific/failable.py
```diff
@@ -0,0 +1,165 @@
...
+    def swap(
+        self: _DiverseFailableType,
+    ) -> KindN[_DiverseFailableType, _SecondType, _FirstType, _ThirdType]:
+        """Swaps value and error types in ``DiverseFailableN``."""
+
+    @classmethod
+    @abstractmethod
+    def from_failure(
+        cls: Type[_DiverseFailableType],
+        inner_value: _SecondType,
```
sobolevn, line 156: This should be `inner_value: _UpdatedType`
Because otherwise we would have type inference bugs.
Because `_SecondType` is a generic-bound type in this context.
And we need a free one.

sobolevn: https://github.com/dry-python/returns/blob/master/returns/interfaces/applicative.py#L157

typesafety/test_interfaces/test_specific/test_failable/test_diverse_failable.yml
```diff
@@ -0,0 +1,145 @@
...
+        SupportsKind3['MyClass', _FirstType, _SecondType, _ThirdType],
+        DiverseFailable3[_FirstType, _SecondType, _ThirdType],
+    ):
+        def swap(self) -> MyClass[_FirstType, _SecondType, _ThirdType]:
+            ...
+
+        @classmethod
+        def from_failure(
+            cls,
+            inner_value: _FirstType
```
sobolevn, line 135: Try `inner_value: _SecondType` here, it should be incorrect.

sobolevn (changes requested):
Thanks a lot! Great work!

Some extras:
- [ ] We can refactor `cond` to use `DiverseFailable` https://github.com/dry-python/returns/blob/master/returns/methods/cond.py#L12
- [ ] We can refactor `unify` to use `DiverseFailable` https://github.com/dry-python/returns/blob/master/returns/pointfree/unify.py

returns/interfaces/specific/failable.py
```diff
@@ -0,0 +1,165 @@
...
+        function: Callable[[_SecondType], _NewFirstType],
+    ) -> None:
+        """Ensures that you cannot alt a success."""
+        assert_equal(
+            container.from_value(raw_value),
+            container.from_value(raw_value).alt(function),
+        )
+
+
+class FailableN(
```
sobolevn, line 106: One more thing. I really dislike current `Maybe` situation. It has no way to directly access a failure case from its class. And we have to import `Nothing`.

`DiverseFailable` has `.from_failure`, `Failable` should have `empty` property instead. So, we would be able to add `empty: ClassVar[Maybe[Any]] = Nothing` to `Maybe`!

sobolevn (changes requested):

returns/interfaces/specific/failable.py
thepabloaguilar: Is this worth it?? Because `empty` is `Maybe` specific behavior, if I have a class like:
```python
class Number(Failable[_ValueType, float('nan')]):
    ...

n: Number[10.0]
print(n.empty) # => nan
```

Does not make much sense to me!

sobolevn: Well, this is kind of symmetric to what we have in `Result`. And `empty` can later be used with `monoid` if we would ever need one.

thepabloaguilar: Seems weird when we think about `Result`, like, what should `Result.empty` means?
Can you provide a practical example without using `Maybe`??

sobolevn: Oh, I see. `Failable` would require `Result` to also have this prop. 
Let's discard this idea for now.

typesafety/test_interfaces/test_specific/test_failable/test_diverse_failable.yml
thepabloaguilar: Done!

returns/interfaces/specific/failable.py
thepabloaguilar: Done!

returns/interfaces/specific/maybe.py
thepabloaguilar: Done!

docs/pages/interfaces.rst
thepabloaguilar: Done!

docs/pages/interfaces.rst
```diff
@@ -447,6 +447,15 @@ SupportsEquality
   :members:
   :private-members:
 
+Failable
```
sobolevn, line 450: `Failable` should go after `Container`

We use semantic ordering here.

returns/interfaces/failable.py
```diff
@@ -0,0 +1,166 @@
...
+    ) -> None:
+        """Ensures that you cannot rescue a success."""
+        assert_equal(
+            container.from_value(raw_value),
+            container.from_value(raw_value).rescue(function),
+        )
+
+
+@final
+class _DiverseFailableLawSpec(LawSpecDef):
```
sobolevn, line 55: Looks like we should only have `alt_short_circuit_law` in `_DiverseFailableLawSpec`.
All other laws can be moved to `_FailableLawSpec`

returns/interfaces/failable.py
```diff
@@ -0,0 +1,166 @@
...
+        function: Callable[[_SecondType], _NewFirstType],
+    ) -> None:
+        """Ensures that you cannot alt a success."""
+        assert_equal(
+            container.from_value(raw_value),
+            container.from_value(raw_value).alt(function),
+        )
+
+
+class FailableN(
```
sobolevn, line 107: We also tend to sctructure interface files differently: 
- LawsSpec
- Its interface
- Another LawsSpec
- Another Interface

sobolevn (changes requested):
Great work! 

Just some minor polishing and we are ready to go!

returns/interfaces/failable.py
thepabloaguilar: I think not, because `map_short_circuit_law` and `bind_short_circuit_law` use `.from_failure` method.
The `.from_failure` is only available on `DiverseFailableN`!

sobolevn: Then we come back to `.empty` feature again. Because, `.map` should not work on `Nothing` or `NaN` or anything else. And we need to somehow express it.

Maybe we can create a new interfaces called `SingleFailableN`?
In this case we would have:

```
FailableN -> SingleFailableN -> Maybe
|
 \
  \
    - - - -> DiverseFailableN -> Result
```

And `SingleFailableN` can have `empty` property with `Nothing` for `Maybe`.
This way, we would be able to add laws for it as well.

What do you think?

thepabloaguilar: What are you proposing here?

Using the current scenario we'll have something like this:
```python
class FailableN(ContainerN[...], RescuableN[...]):
    ...

class SingleFailableN(FailableN[...]):
    empty: ClassVar[Any]

class DiverseFailableN(FailableN[...], BiMappableN[...]):
    ...
```

But, we won't able to move the tests to the `_FailableLawSpec` because `.from_value` have to stay on `DiverseFailableN`!

sobolevn: Yes, they will have two different tests:

```python
class SingleFailableLawSpec(LawSpecDef):
    @law_definition
    def map_short_circuit_law(
        raw_value: _SecondType,
        container: 'DiverseFailableN[_FirstType, _SecondType, _ThirdType]',
        function: Callable[[_FirstType], _NewFirstType],
    ) -> None:
        """Ensures that you cannot map a failure."""
        assert_equal(
            container.empty,
            container.empty.map(function),
        )

    # etc for all other laws
```

returns/interfaces/failable.py
```diff
@@ -0,0 +1,230 @@
...
+    Like ``Maybe`` types where the only failed value is ``Nothing``.
+    """
+
+    _laws: ClassVar[Sequence[Law]] = (
+        Law2(_SingleFailableLawSpec.map_short_circuit_law),
+        Law2(_SingleFailableLawSpec.bind_short_circuit_law),
+    )
+
+    #: Alias for failed value
+    empty: ClassVar['SingleFailableN[_FirstType, _SecondType, _ThirdType]']
```
sobolevn, line 131: Let's make it an abstract property, like here: https://github.com/dry-python/returns/blob/master/returns/interfaces/specific/reader.py#L95

So, you will be forced to override it inside yout subtypes.

returns/interfaces/failable.py
```diff
@@ -0,0 +1,230 @@
...
+        function: Callable[[_FirstType], _NewFirstType],
+    ) -> None:
+        """Ensures that you cannot map a failure."""
+        assert_equal(
+            container.from_failure(raw_value),
+            container.from_failure(raw_value).map(function),
+        )
+
+    @law_definition
+    def bind_short_circuit_law(
```
sobolevn, line 163: Currently we have only `bind`, `map`, `alt`, and `rescue` covered. But, `apply` should also short circuit for `Failable` subtypes.

Let's add `.apply` as well. To both interfaces.

returns/maybe.py
```diff
@@ -420,6 +423,7 @@ def failure(self):
 
 Maybe.success_type = _Some
 Maybe.failure_type = _Nothing
+Maybe.empty = _Nothing()
```
sobolevn, line 426: We have to use `Nothing` here. Because it is a singleton.

sobolevn (changes requested):
Awesome! There are some last thing to fix and we are ready.

I also had this idea while reading your PR: https://github.com/dry-python/returns/issues/634

returns/interfaces/failable.py
thepabloaguilar: Ok, I made. I got two errors with `IOResult` with apply when the law was:
```python
class Laws:
    @law_definition
    def apply_short_circuit_law(
        raw_value: _SecondType,
        container: 'DiverseFailableN[_FirstType, _SecondType, _ThirdType]',
        wraped_function: KindN[
            'DiverseFailableN',
            Callable[[_FirstType], _NewFirstType],
            _SecondType,
            _ThirdType,
        ],
    ) -> None:
        """Ensures that you cannot apply a failure."""
        assert_equal(
            container.from_failure(raw_value),
            container.from_failure(raw_value).apply(wraped_function),
        )
```

I modified to:
```python
class Laws:
    @law_definition
    def apply_short_circuit_law(
        raw_value: _SecondType,
        container: 'DiverseFailableN[_FirstType, _SecondType, _ThirdType]',
        function: Callable[[_FirstType], _NewFirstType],
    ) -> None:
        """Ensures that you cannot apply a failure."""
        assert_equal(
            container.from_failure(raw_value),
            container.from_failure(raw_value).apply(container.from_value(function)),
        )
```
And now it's working!!

typesafety/test_interfaces/test_failable/test_single_failable.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    class MyClass(
+        SupportsKind3['MyClass', _FirstType, _SecondType, _ThirdType],
+        SingleFailable3[_FirstType, _SecondType, _ThirdType],
+    ):
+        ...
+
+
+- case: single_failable_inheritance_missing
+  disable_cache: true
```
sobolevn, line 37: Please, check that all `disable_cache: false` are set in new tests.

typesafety/test_interfaces/test_failable/test_single_failable.yml
```diff
@@ -0,0 +1,54 @@
...
+    _SecondType = TypeVar('_SecondType')
+
+    @final
+    class MyClass(
+        SupportsKind2['MyClass', _FirstType, _SecondType],
+        SingleFailable2[_FirstType, _SecondType],
+    ):
+        ...
+  out: |
+    main:10: error: Final class main.MyClass has abstract attributes "apply", "bind", "from_iterable", "from_value", "map", "rescue"
```
sobolevn, line 54: Why don't we have `empty` prop here? 🤔

returns/methods/cond.py
```diff
@@ -1,22 +1,22 @@
...
 
 
 def internal_cond(
-    container_type: Type[_ResultKind],
+    container_type: Type[_DiverseFailableKind],
     is_success: bool,
     success_value: _ValueType,
     error_value: _ErrorType,
-) -> Kind2[_ResultKind, _ValueType, _ErrorType]:
+) -> Kind2[_DiverseFailableKind, _ValueType, _ErrorType]:
```
sobolevn, line 17: Looks like this is not 100% related, but.

Let's use `KindN` here. Because I can see a bug in `Kind2` here. 
Here's a failing test:

```yml
- case: unify_reader_future_result
  disable_cache: false
  main: |
    from returns.pointfree import unify
    from returns.context import ReaderFutureResult

    def test(arg: int) -> ReaderFutureResult[int, bool, float]:
        ...

    x: ReaderFutureResult[int, str, float]
    reveal_type(unify(test)(x))  # N: Revealed type is 'returns.context.requires_context_future_result.RequiresContextFutureResult[builtins.int, Union[builtins.str*, builtins.bool], builtins.float]'
```

sobolevn: You can just create a new issue out of it for now.

sobolevn (changes requested):
👍

typesafety/test_interfaces/test_failable/test_single_failable.yml
thepabloaguilar: This is an interesting question, I tried to add `metaclass=ABCMeta` or inherit directly from `ABC` but nothing make this fails!
This code is working without any issue:
```python
>>> from typing import TypeVar
>>> from typing_extensions import final
>>> from returns.interfaces.failable import SingleFailable2
>>> from returns.primitives.hkt import SupportsKind2
>>> 
>>> _FirstType = TypeVar('_FirstType')
>>> _SecondType = TypeVar('_SecondType')
>>> @final
... class MyClass(
...     SupportsKind2['MyClass', _FirstType, _SecondType],
...     SingleFailable2[_FirstType, _SecondType],
... ):
...     ...
... 
>>> MyClass()
```

thepabloaguilar: I can't figure out why this is not failing

thepabloaguilar: Do you have some idea?

returns/methods/cond.py
thepabloaguilar: https://github.com/dry-python/returns/issues/637

returns/interfaces/failable.py
```diff
@@ -0,0 +1,259 @@
...
+
+    _laws: ClassVar[Sequence[Law]] = (
+        Law2(_SingleFailableLawSpec.map_short_circuit_law),
+        Law2(_SingleFailableLawSpec.bind_short_circuit_law),
+        Law2(_SingleFailableLawSpec.apply_short_circuit_law),
+    )
+
+    @property
+    @abstractmethod
+    def empty(self) -> 'SingleFailableN[_FirstType, _SecondType, _ThirdType]':
```
sobolevn, line 145: We need to annotate `self` here.

typesafety/test_interfaces/test_failable/test_single_failable.yml
sobolevn: `ReaderLike2` works correctly: https://github.com/dry-python/returns/blob/master/typesafety/test_interfaces/test_specific/test_reader/test_reader_like2.yml#L70 It has abstract property `.empty`

The only difference is `self` annotation.

sobolevn (commented):
🔝

returns/interfaces/failable.py
sobolevn: https://github.com/dry-python/returns/blob/master/returns/interfaces/specific/reader.py#L97

thepabloaguilar: It didn't work 😢 😢

```python
class SingleFailableN(
    FailableN[_FirstType, _SecondType, _ThirdType],
    Lawful['SingleFailableN[_FirstType, _SecondType, _ThirdType]'],
):
    @property
    @abstractmethod
    def empty(
        self: _SingleFailableType,
    ) -> KindN[_SingleFailableType, _FirstType, _SecondType, _ThirdType]:
        """This property represents the failed value."""
```

thepabloaguilar:
@sobolevn I  found the "error"!! And OMG was a big mistake hahaha

I used the wrong type to make the `SingleFailableN` type alias:
```python
#: Type alias for kinds with two types arguments.
SingleFailable2 = FailableN[_FirstType, _SecondType, NoReturn]

#: Type alias for kinds with three type arguments.
SingleFailable3 = FailableN[_FirstType, _SecondType, _ThirdType]
```

🙈 🙈

thepabloaguilar:
Btw, I prefer to fix #634 in a separated PR. I think this one is big enough!

sobolevn (approved):

PR #648, opened 2020-09-29
Modifies `cond` and `unify` to use `KindN` instead of `Kind2`

### LINKED ISSUE #637: `cond` and `unify` should use `KindN` instead of `Kind2`
# Bug report

## What's wrong

We can have some problems when try to work with containers Kind3!

To verify if everything is working correctly we need to add this test case to `cond` typesafety tests:
```yml
- case: unify_reader_future_result
  disable_cache: false
  main: |
    from returns.pointfree import unify
    from returns.context import ReaderFutureResult

    def test(arg: int) -> ReaderFutureResult[int, bool, float]:
        ...

    x: ReaderFutureResult[int, str, float]
    reveal_type(unify(test)(x))  # N: Revealed type is 'returns.context.requires_context_future_result.RequiresContextFutureResult[builtins.int, Union[builtins.str*, builtins.bool], builtins.float]'
```

## How is that should be

We just have to change from `Kind2` to `KindN`

```python
def internal_cond(
    container_type: Type[_DiverseFailableKind],
    is_success: bool,
    success_value: _ValueType,
    error_value: _ErrorType,
) -> KindN[_DiverseFailableKind, _ValueType, _ErrorType, _ThirdType]:
    ...
```

thepabloaguilar:
# Modifies `cond` and `unify` to use `KindN` instead of `Kind2`

## Checklist

## Related issues

Closes #637

### DIFF
--- a/returns/methods/cond.py
+++ b/returns/methods/cond.py
@@ -1,7 +1,8 @@
 from typing import Type, TypeVar
 
+from returns.context import NoDeps
 from returns.interfaces.failable import DiverseFailableN
-from returns.primitives.hkt import Kind2, kinded
+from returns.primitives.hkt import KindN, kinded
 
 _ValueType = TypeVar('_ValueType')
 _ErrorType = TypeVar('_ErrorType')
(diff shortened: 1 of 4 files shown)


sobolevn (approved):

sobolevn:
Thanks a lot!

PR #652, opened 2020-10-02
Add a hypothesis test case with an expected failure

### LINKED ISSUE #630: Add a hypothesis test case with an expected failure
We need a test like https://github.com/dry-python/returns/blob/master/tests/test_contrib/test_hypothesis/test_laws/test_custom_type_with_init.py but it should fail.

This way we would check that it is possible to violate the law and your test will fail.
We would also need to check that we have `strict-xfail` setup in `pytest`.

@thepabloaguilar will you take this one, please?

thepabloaguilar:
# Add a hypothesis test case with an expected failure

@sobolevn this is what are you expecting?

## Checklist

## Related issues

Closes #630

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -96,6 +96,9 @@ strictness = long
 # ignores some directories:
 norecursedirs = *.egg .eggs dist build docs .tox .git __pycache__
 
+# Active the strict mode of xfail
+xfail_strict = true
+
 # Adds these options to each `pytest` run:
 addopts =
   --doctest-modules
--- a/tests/test_contrib/test_hypothesis/test_laws/test_wrong_custom_type_with_init.py
+++ b/tests/test_contrib/test_hypothesis/test_laws/test_wrong_custom_type_with_init.py
@@ -1,0 +1,35 @@
+from typing import Callable, TypeVar
+
+import pytest
+
+from returns.contrib.hypothesis.laws import check_all_laws
+from returns.interfaces import mappable
+from returns.primitives.container import BaseContainer
+from returns.primitives.hkt import SupportsKind1
+
+pytestmark = pytest.mark.xfail
+
+_ValueType = TypeVar('_ValueType')
+_NewValueType = TypeVar('_NewValueType')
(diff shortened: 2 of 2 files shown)


returns/contrib/hypothesis/laws.py
```diff
@@ -28,13 +28,15 @@ class _Settings(NamedTuple):
...
     use_init: bool
+    mark_xfail: bool
 
 
 def check_all_laws(
     container_type: Type[Lawful],
     *,
     settings_kwargs: Optional[Dict[str, Any]] = None,
     use_init: bool = False,
+    mark_xfail: bool = False,
```
sobolevn, line 39: Nope, sorry.

I was expecting that `xfail` should be a mark of the whole test module, like so:

```python
pytestmark = pytest.mark.xfail(...)

check_all_laws(...)
```

returns/contrib/hypothesis/laws.py
```diff
@@ -28,13 +28,15 @@ class _Settings(NamedTuple):
 
     settings_kwargs: Dict[str, Any]
     use_init: bool
+    mark_xfail: bool
```
sobolevn, line 31: I really don't want to add this flag, because I don't see any reasons why anyone would need it. Do you have some use-cases in mind?

sobolevn (changes requested):

sobolevn:
Thanks a lot for your work! Sorry, I was not clear enough with this task 😞

returns/contrib/hypothesis/laws.py
thepabloaguilar: Oh, now I got it hahaha

returns/contrib/hypothesis/laws.py
thepabloaguilar: No, it's because I didn't read yet about pytest module marker!! I'll revert it

sobolevn (approved):

sobolevn:
🎉

PR #653, opened 2020-10-02
Allows `cond` HKT method & pointfree to work with `SingleFailableN`

### LINKED ISSUE #634: Allow cond to work with `SingleFailableN`
We can probably try to make a special overload for `cond` and `SingleFailableN`.
That's how it should work:

```python
@overload
def cond(typ: Type[_SingleFailableKind], predicate: Callable[[_FirstType], bool], first: _FirstType) -> KindN[...]: ...
```

And we would have the second overload: the way it is defined in #624 
So, it would work for both types.

thepabloaguilar:
# Allows `cond` HKT method & pointfree to work with `SingleFailableN`

## Checklist

## Related issues

Closes #634

### DIFF
--- a/docs/pages/pointfree.rst
+++ b/docs/pages/pointfree.rst
@@ -218,8 +218,9 @@ kind of manipulation.
 cond
 ----
 
-Sometimes we need to create ``DiverseFailableN`` containers
-(e.g. ``ResultLikeN``) based on a boolean expression, ``cond`` can help us.
+Sometimes we need to create ``SingleFailableN`` or ``DiverseFailableN``
+containers (e.g. ``Maybe``, ``ResultLikeN``) based on a boolean expression,
+``cond`` can help us.
 
 See the example below:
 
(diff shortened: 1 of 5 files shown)


docs/pages/pointfree.rst
```diff
@@ -218,16 +218,17 @@ kind of manipulation.
 cond
 ----
 
-Sometimes we need to create ``DiverseFailableN`` containers
-(e.g. ``ResultLikeN``) based on a boolean expression, ``cond`` can help us.
+Sometimes we need to create ``SingleFailableN`` or ``DiverseFailableN``
+containers (e.g. ``Maybe``, ``ResultLikeN``) based on a boolean expression,
+``cond`` can help us.
```
sobolevn, line 223: Let add a new sentence here:

```
Consider ``cond`` to be a functional ``if``.
```

sobolevn, line 223:
```suggestion
``cond`` can help us.

Consider ``cond`` to be a functional ``if``.
```

sobolevn (approved):

sobolevn:
Thanks a lot! Nice one!

PR #654, opened 2020-10-02
Added __repr__ in BaseContainer class

### LINKED ISSUE #636: Consider implementing __str__ / __repr__ for suitable containers
# Bug report

## What's wrong

When writing test cases for a method that returns `Maybe[T]`, I found that pytest reports a failure as follows which is not very readable:
> <returns.maybe._Some object at 0x7fddfe56c140> != <returns.maybe._Some object at 0x7fddfe5aea00>

## How is that should be

It would be much better, if `Some` implemented `__str__` (and maybe also `__repr__`) so that it could look like below:
> <Some(10) != Some(20)>

## System information

- `python` version: 3.8
- `returns` version: 0.14.0

raoashish10:
# I have made things!

## Checklist

## Related issues

- Closes #636 
Added the __ repr __() to the BaseContainer class. I haven't removed the __ str __(), I  have just kept both. Let me know if any changes are required.

### DIFF
--- a/returns/primitives/container.py
+++ b/returns/primitives/container.py
@@ -22,20 +22,13 @@ class BaseContainer(Immutable, metaclass=ABCMeta):
         """
         object.__setattr__(self, '_inner_value', inner_value)  # noqa: WPS609
 
-    def __str__(self) -> str:
+    def __repr__(self) -> str:
         """Converts to string."""
         return '<{0}: {1}>'.format(
             self.__class__.__qualname__.strip('_'),
             str(self._inner_value),
         )
 
-    def __repr__(self) -> str:
-        """Used to display details of object"""
-        return '<{0}: {1}>'.format(
-            self.__class__.__qualname__.strip('_'),
-            str(self._inner_value),
-        )
-        
     def __eq__(self, other: Any) -> bool:
         """Used to compare two 'Container' objects."""
         return container_equality(self, other)  # type: ignore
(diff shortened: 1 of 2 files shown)


returns/primitives/container.py
```diff
@@ -29,6 +29,13 @@ def __str__(self) -> str:
             str(self._inner_value),
         )
 
+    def __repr__(self) -> str:
```
sobolevn, line 32: There's no need to add a new method, we can just rename `__str__` to `__repr__`

sobolevn (changes requested):
Thanks a lot! 

Please, see what tests will fail and fix them as well.

returns/primitives/container.py
raoashish10: Sure I will do that!

returns/primitives/container.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, inner_value) -> None:
         """
         object.__setattr__(self, '_inner_value', inner_value)  # noqa: WPS609
 
-    def __str__(self) -> str:
+    def __repr__(self) -> str:
         """Converts to string."""
```
thepabloaguilar, line 26: Can you update the docstring, please?
`__repr__` is more than a simple string conversion.

raoashish10: Yes I will change it give me a minute. I had kept it as """Used to display details of the object""" before. Is that fine?

sobolevn: 👍

sobolevn (approved):

sobolevn:
Thanks a lot! 👍

raoashish10:
@sobolevn I look to contribute to bigger issues in the future!

sobolevn:
@raoashish10 feel free to choose any other issues from the backlog.

PR #659, opened 2020-10-03
Bumps poetry version alone

sobolevn:


sobolevn:
Ok, here's why:

```
E     main:1: error: Cannot find implementation or library stub for module named 'returns.example' (diff)
E     main:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports (diff)
E     main:3: note: Revealed type is 'Any' (diff)
```

sobolevn:
So, this happens because of how `poetry` now handles local installs: 

> /Users/sobolev/Documents/github/returns/.venv/lib/python3.7/site-packages/returns.pth

PR #669, opened 2020-10-10
Sets documentation theme color attribute

### LINKED ISSUE #649: Bad highlight of keyword on documentation
# Bug report

## What's wrong

When we search something in the `returns` documentation the highlighted term is hard to see on the page. See the example below, this is the `Result` container page and I searched by `result` keyword:

image

URL to reproduce: https://returns.readthedocs.io/en/latest/pages/result.html?highlight=result

## How is that should be

People has to see the highlighted keyword, working example:

image

I searched by `logging` keyword!
URL: https://eliot.readthedocs.io/en/stable/generating/actions.html?highlight=logging#logging-functions

thepabloaguilar:
# Sets documentation theme color attribute

After some research I found where we need to modify!!
It's just a property called `color` we have to add in our `conf.py`.
[Reference](https://sphinx-typlog-theme.readthedocs.io/en/latest/options.html#color)

It's working:
image

I've used `#E8371A`, it's the same from `sphinx-typlog-theme`. This color is used to:
* Keyword highlight
* Link hover
* Mouse selection

You can try it online [here](https://returns--669.org.readthedocs.build/en/669/)

Btw, the Bindable's Laws was in a wrong position!!

## Checklist

## Related issues

Closes #649

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -165,6 +165,7 @@ html_theme_options = {
     ),
     'github_user': 'dry-python',
     'github_repo': 'returns',
+    'color': '#E8371A',
 }
 
 # Add any paths that contain custom static files (such as style sheets) here,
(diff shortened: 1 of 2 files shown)


sobolevn (approved):
It was easy! I though we would have to fork it / use a custom css file 😆

PR #677, opened 2020-10-10
Test that `Fold` methods are recursion safe

### LINKED ISSUE #644: Test that `Fold` methods are recursion safe
Big iterables might potentially trigger `RecursionError` if we would have some wrong abtractions inside.
So, we need to tests that even iterables with big number of elements (bigger than recursion limit) do not raise an error.

We need to test:
- Public methods of `Fold` here https://github.com/dry-python/returns/blob/master/returns/iterables.py
- Recusion limit: https://docs.python.org/3/library/sys.html#sys.getrecursionlimit

Related https://github.com/dry-python/returns/pull/626
Related #598

timqsh:
closes #644

### DIFF
--- a/returns/iterables.py
+++ b/returns/iterables.py
@@ -25,7 +25,7 @@ class AbstractFold(object):
     .. rubric:: Implementation
 
     ``AbstractFold`` and ``Fold`` types are special.
-    They have double definition for all method: public and protected ones.
+    They have double definition for each method: public and protected ones.
     Why?
 
     Because you cannot override ``@kinded`` method due to a ``mypy`` bug.
--- a/returns/iterables.py
+++ b/returns/iterables.py
@@ -25,7 +25,7 @@ class AbstractFold(object):
     .. rubric:: Implementation
 
     ``AbstractFold`` and ``Fold`` types are special.
-    They have duoble definition for all method: public and protected ones.
+    They have double definition for all method: public and protected ones.
     Why?
 
     Because you cannot override ``@kinded`` method due to a ``mypy`` bug.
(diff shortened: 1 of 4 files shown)


tests/test_iterables/test_fold/test_collect.py
```diff
@@ -200,3 +201,11 @@ async def test_fold_collect_future_result(subtests):
             assert await Fold.collect(
                 iterable, sequence.from_value(()),
             ) == await sequence
+
+
+def test_fold_collect_recursion_limit():
+    """Ensures that ``.collect`` method is recurion safe."""
+    limit = sys.getrecursionlimit() + 1
+    iterable = [IO(1) for _ in range(limit)]
```
thepabloaguilar, line 209: Maybe it would be better to use a generator here!
What do you think @sobolevn?

returns/iterables.py
```diff
@@ -25,7 +25,7 @@ class AbstractFold(object):
     .. rubric:: Implementation
 
     ``AbstractFold`` and ``Fold`` types are special.
-    They have duoble definition for all method: public and protected ones.
+    They have double definition for all method: public and protected ones.
```
sobolevn, line 28: for all methods

sobolevn: Or `for each method` 🤔

tests/test_iterables/test_fold/test_collect.py
sobolevn: 👍

sobolevn (changes requested):
Thanks a lot! A little polishing and we are good to go!

tests/test_iterables/test_fold/test_collect.py
timqsh: I like that suggestion. Done.

returns/iterables.py
timqsh: For each sounds better to me

thepabloaguilar (approved):

thepabloaguilar:
Great work!! 🎉🎉

PR #717, opened 2020-11-06
Makes `_Nothing` a singleton

### LINKED ISSUE #693: Make `_Nothing.__new__` a real singleton
Currenlty, it is still possible to produce another `Nothing` instance.
Example: 

```python
NewNothing = _Nothing()
```

https://github.com/tek/amino/blob/master/amino/maybe.py#L163-L168

thepabloaguilar:
# Makes `_Nothing` a singleton

## Checklist

## Related issues

Closes #693

### DIFF
--- a/returns/maybe.py
+++ b/returns/maybe.py
@@ -285,6 +285,12 @@ class _Nothing(Maybe[Any]):
     """Represents an empty state."""
 
     _inner_value: None
+    _instance: Optional['_Nothing'] = None
+
+    def __new__(cls, *args: Any, **kwargs: Any) -> '_Nothing':
+        if cls._instance is None:
+            cls._instance = object.__new__(cls)  # noqa: WPS609
+        return cls._instance
 
     def __init__(self, inner_value: None = None) -> None:  # noqa: WPS632
         """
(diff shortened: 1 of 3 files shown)


returns/maybe.py
```diff
@@ -285,6 +285,12 @@ class _Nothing(Maybe[Any]):
     """Represents an empty state."""
 
     _inner_value: None
+    _instance: Optional['_Nothing'] = None
+
+    def __new__(cls, *args: Any, **kwargs: Any) -> '_Nothing':
+        if _Nothing._instance is None:  # noqa: WPS437
```
sobolevn, line 291: Shouldn't it be `cls._instance`?

sobolevn:
Thanks a lot! 👍

returns/maybe.py
thepabloaguilar: Yes, it's better!!

returns/maybe.py
```diff
@@ -285,6 +285,12 @@ class _Nothing(Maybe[Any]):
     """Represents an empty state."""
 
     _inner_value: None
+    _instance: Optional['_Nothing'] = None
+
+    def __new__(cls, *args: Any, **kwargs: Any) -> '_Nothing':
+        if cls._instance is None:  # noqa: WPS437
```
sobolevn, line 291: Do we still need `noqa` here?

thepabloaguilar: Nop

sobolevn (approved):

sobolevn:
Thanks!

Issue #862, opened 2021-03-14
returns.assert_trace doesn't work with @safe-wrapped functions


ftruzzi:
First of all, thank you! I really appreciate your work and your help on the Telegram group :)

# Bug report

## What's wrong
`returns.assert_trace` fails to "intercept" containers created from `@safe` decorators.

```python
# works
def func_1(n) -> Result[int, Exception]:
    if n == 0:
        return Failure(Exception())
    return Success(n + 1)

# doesn't work
@safe
def func_1(n) -> int:
    if n == 0:
        raise Exception()
    return n + 1


def test_func_1(returns: ReturnsAsserts):
    with returns.assert_trace(Success, func_1):
        flow(1, func_1)
```

The test with the wrapped function returns

```
FAILED returns_test.py::test_func_1 - Failed: No container Success was created
```

I tried doing this, but no difference:
```python
def test_func_1(returns: ReturnsAsserts):
    with returns.assert_trace(Success, safe(func_1)):
        flow(1, safe(func_1))

```

## How is that should be
The above test should pass

## System information

- `python` version: 3.8.0
- `returns` version: 0.15.0
- `pytest` version (if any): 6.2.2

thepabloaguilar:
Hello @ftruzzi, thank you for open this issue!

I found the problem here, our plugin can search the container in the entire call stack, so, the test below should pass:
```python
def func_one():
    return func_two()

def func_two():
    return Success(42)

def test_something(returns: ReturnsAsserts):
    with returns.assert_trace(Success, func_one):
        func_one()
```

The problem here is when we use `safe`, we don't have `func_one` in the call stack anymore, let me try to explain:
```python
def func_one():
    return Success(42)

# When we found `Success` call while calling `func_one` the call stack is something like:
# - Success (call)
# - func_one (call)
# So we can reach `func_one`

@safe
def func_one():
    return 42

# Using `safe` decorator all of the containers are created inside it
# So, when we find a desired container creation call we have this:
# - Success (call)
# - safe (call)
# Instead of
# - func_one (call)
# - Success (call)
# - func_one (call)
# And that makes sense because we already returned from `func_one`
```

ftruzzi:
Thanks for the explanation!

I have never looked into the Python call stack, but maybe this can help?

https://medium.com/@genexu/extracting-the-wrapped-function-from-a-decorator-on-python-call-stack-2ee2e48cdd8e

thepabloaguilar:
> Thanks for the explanation!
[...]

PR #863, opened 2021-03-15
make assert_trace work with `@safe` (#862)

ftruzzi:
# I have made things!

## Checklist

## Related issues

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,10 +3,6 @@ name: test
 'on':
   push:
   pull_request:
-    branches:
-      # Branches from forks have the form 'user:branch-name'. Reference:
-      # https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662/9
-      - '**:**'
   workflow_dispatch:
 
 jobs:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ See [0Ver](https://0ver.org/).
 
 - Makes `_Nothing` a singleton
 - Fixes typos in documentation
+- Fixes that `assert_trace` was not catching containers from `@safe`-wrapped functions
 
 
 ## 0.15.0 aka The initial HKT release
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@ See [0Ver](https://0ver.org/).
 
 - Makes `_Nothing` a singleton
 - Fixes typos in documentation
+- Refactor `flow` function
 - Fixes that `assert_trace` was not catching containers from `@safe`-wrapped functions
-
 
 ## 0.15.0 aka The initial HKT release
 
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -15,4 +15,4 @@ pygments==2.8.1
 typing-extensions==3.7.4.3
 mypy==0.812
 pytest==6.2.2
-hypothesis==6.8.0
+hypothesis==6.8.1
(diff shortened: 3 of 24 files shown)


sobolevn (approved):
Thanks a lot!

thepabloaguilar (commented):
Sorry for the delay!!

This PR resolves the problem, but I think we're putting some dirty things from our pytest plugin inside the `core`.
Do you mind any other solution?

ftruzzi:
@thepabloaguilar thanks making me think a bit more about it!

How about moving `_DesiredFunctionFound` back to the plugin and making it a subclass of `BaseException` instead? Should not be caught by `@safe` this way and it's definitely a cleaner fix.

sobolevn:
I am not the codeowner for this feature, but I like it! 👍

thepabloaguilar:
@ftruzzi I have another idea that I'll be glad if you could test it before!!
Let's move our `pytest.fail(...)` from [here](https://github.com/dry-python/returns/blob/master/returns/contrib/pytest/plugin.py#L70) to [here](https://github.com/dry-python/returns/blob/master/returns/contrib/pytest/plugin.py#L138)!

Our trace function will be something like this:
```python
def _trace_function(
    trace_type: _ReturnsResultType,
    function_to_search: _FunctionType,
    frame: FrameType,
    event: str,
    arg: Any,
) -> None:
    is_desired_type_call = (
        event == 'call' and frame.f_code is trace_type.__code__
    )
    if is_desired_type_call:
        current_call_stack = inspect.stack()
        function_to_search_code = getattr(function_to_search, '__code__', None)
        for frame_info in current_call_stack:
            if function_to_search_code is frame_info.frame.f_code:
                pytest.fail(...)
                return
```

ftruzzi:
Sorry @thepabloaguilar I think I'm a bit confused -- shouldn't the test **pass** if the `function_to_search_code is frame_info.frame.f_code` condition is satisfied?

thepabloaguilar:
@ftruzzi you're totally right, my bad, it should be:

```python
def _trace_function(
    trace_type: _ReturnsResultType,
    function_to_search: _FunctionType,
    frame: FrameType,
    event: str,
    arg: Any,
) -> None:
    is_desired_type_call = (
        event == 'call' and frame.f_code is trace_type.__code__
    )
    if is_desired_type_call:
        current_call_stack = inspect.stack()
        function_to_search_code = getattr(function_to_search, '__code__', None)
        for frame_info in current_call_stack:
            if function_to_search_code is frame_info.frame.f_code:
                return
        pytest.fail(...)
```

thepabloaguilar:
Well, we can't do what I've suggested! I remembered why hahaha
We can't say from inside the `_trace_function` if we did not find the desired function 'cause we are not on top of all the posible calls since `_trace_function` is called every time that a call for any function is made!

Just let me illustrate my explanation:

```python
# CALL STACK
# Call 6
# Call 5
* Call 4 -> `_trace_function` is called, we can't say here if the desired function was not found since there will be more two calls
* Call 3
* Call 2
* Call 1
```

thepabloaguilar:
Please @ftruzzi continue with your solution, if @sobolevn doesn't see any problem with modifying `@safe` to not handle our "internal" exceptions!

sobolevn:
I like this solution https://github.com/dry-python/returns/pull/863#issuecomment-800944015

sobolevn (approved):
Thanks!

ftruzzi:
Thank you both @sobolevn and @thepabloaguilar, it's nice to learn new stuff and contribute :)

PR #871, opened 2021-03-18
Closes #706 refactor `flow` function

### LINKED ISSUE #706: Measure and possibly refactor `flow` function
Maybe it is better to use `for` loop inside a `flow`'s body? Let's measure.

`flow` is a hot path, so we need to speed this up as much as we can.
Link: https://github.com/dry-python/returns/blob/0f7d02d0c491a7c65c74e6c0645f12fccc53fe18/returns/_internal/pipeline/flow.py

novikovfred:
Hi @sobolevn 

it's only try to close #706 

I did some experimentation with this where I refactored flow to:
``````
def flow_faster(
    instance: _InstanceType,
    *functions: _PipelineStepType,
) -> _ReturnType:
    return reduce(lambda composed, function: function(composed), functions, instance)
``````

**I made some benchmarks:**
``````
import timeit

setup='''
from returns.curry import curry
from returns._internal.pipeline.flow import flow, flow_for, flow_faster
from returns.pointfree import bind
from returns.result import Success, Failure

def regular_function(arg):
    return arg

def returns_container(arg):
    return Success(arg)

def returns_failure(arg):
    return Failure(arg)

regular_functions = [regular_function for _ in range(300)]
success_container_functions = [bind(returns_container) for _ in range(300)]
failure_container_functions = [bind(returns_failure) for _ in range(300)]
'''

testcode = '''
flow(1, *regular_functions, lambda x: Success(x), *success_container_functions, *failure_container_functions)
'''

print("testcode1 ", timeit.timeit(stmt=testcode, setup=setup, number=100000))

testcode2 = '''
flow_for(1, *regular_functions, lambda x: Success(x), *success_container_functions, *failure_container_functions)
'''
print("testcode2 ", timeit.timeit(stmt=testcode2, setup=setup, number=100000))

testcode3 = '''
flow_faster(1, *regular_functions, lambda x: Success(x), *success_container_functions, *failure_container_functions)
'''
print("testcode3 ", timeit.timeit(stmt=testcode3, setup=setup, number=100000))
``````

Function `flow_for` was taken from #706 

So, I have results like this:
image


sobolevn:
@novikovfred great! Thank you! 👍

sobolevn (approved):

PR #980, opened 2021-06-19
Enables Python 3.10 on CI pipeline

thepabloaguilar:
# Enables Python 3.10 on CI pipeline

This is the first step to start building new features from Python `3.10`, like, Pattern Matching inside `returns`!


.github/workflows/test.yml
```diff
@@ -37,12 +41,14 @@ jobs:
...
     - name: Install dependencies
       run: |
+        poetry config experimental.new-installer false
         poetry config virtualenvs.in-project true
         poetry install
 
         poetry run pip install -U pip
 
     - name: Run tests
+      if: matrix.python-version != '3.10.0-beta.2'
```
sobolevn, line 51: Maybe you can just create a separate step like 

```yml
name: Run lint 
if: matrix.python-version != '3.10.0-beta.2'
run: poetry run flake8 .
```
?

sobolevn (commented):
Thanks! Great work!

I will try to release `wps` soon!

.github/workflows/test.yml
thepabloaguilar: It seems more interesting!

thepabloaguilar: Done!

sobolevn (approved):
Thanks a lot!

PR #981, opened 2021-06-20
Enables Pattern Matching support for `Result` container

thepabloaguilar:
# Enables Pattern Matching support for `Result` container

## Checklist

### DIFF
--- a/docs/pages/result.rst
+++ b/docs/pages/result.rst
@@ -51,34 +51,7 @@ Pattern Matching
 `Structural Pattern Matching <https://www.python.org/dev/peps/pep-0622/>`_,
 see the example below:
 
-.. code:: python
-
-  from returns.result import Success, Failure, safe
-
-  @safe
-  def div(first_number: int, second_number: int) -> int:
-      return first_number // second_number
-
-  match div(1, 0):
-      # Matches if the result stored inside `Success` is `10`
-      case Success(10):
-          print('Result is "10"')
-
-      # Same as above but using match-by-name  and a different value
-      case Success(inner_value=20):
-          print('Result is "20"')
-
-      # Matches any `Success` instance and binds its value to the ``value`` variable
-      case Success(value):
-          print('Result is "{0}"'.format(value))
-
-      # Matches if the result stored inside `Failure` is `ZeroDivisionError`
-      case Failure(ZeroDivisionError):
-          print('"ZeroDivisionError" was raised')
-
-      # Matches any `Failure` instance
-      case Failure(_):
-          print('The division was a failure')
+.. literalinclude:: ../../tests/test_examples/test_result/test_pattern_matching.py
 
 
 Aliases
--- a/docs/pages/result.rst
+++ b/docs/pages/result.rst
@@ -43,6 +43,42 @@ Make sure to check out how to compose container with
 ``flow`` or :ref:`pipe`!
 Read more about them if you want to compose your containers easily.
 
+Pattern Matching
+----------------
+
+``Result`` values can be matched using the new feature of Python 3.10,
+`Structural Pattern Matching <https://www.python.org/dev/peps/pep-0622/>`_,
+see the example below:
+
+.. code:: python
+
+  from returns.result import Success, Failure, safe
+
+  @safe
+  def div(first_number: int, second_number: int) -> int:
+      return first_number // second_number
+
+  match div(1, 0):
+      # Matches if the result stored inside `Success` is `10`
+      case Success(10):
+          print('Result is "10"')
+
+      # Same as above but using match-by-name
+      case Success(inner_value=20):
+          print('Result is "20"')
+
+      # Matches any `Success` instance and binds its value to the ``value`` variable
+      case Success(value):
+          print('Result is "{}"'.format(value))
+
+      # Matches if the result stored inside `Failure` is `ZeroDivisionError`
+      case Failure(ZeroDivisionError):
+          print('"ZeroDivisionError" was raised')
+
+      # Matches any `Failure` instance
+      case Failure(_):
+          print('The division was a failure')
+
 
 Aliases
 -------
@@ -118,40 +154,6 @@ Python's type system does not allow us to do much, so this is required:
 
 Otherwise ``first`` will have ``Result[int, Any]`` type.
 Which is okay in some situations.
-
-What is the difference between ``Success`` and ``_Success``?
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-You might wonder why ``Success`` is a function
-and ``_Success`` is internal type, that should not be used directly.
-
-Well, that's a complicated question. Let's find out.
-
-Let's begin with ``haskell`` definition:
-
-.. code:: haskell
-
-  Prelude> :t Left 1
-  Left 1 :: Num a => Either a b
-  Prelude> :t Right 1
-  Right 1 :: Num b => Either a b
-
-As you can see: ``Left`` (``Failure``) and ``Right`` (``Success``)
-are type constructors: that return ``Either a b`` (``Result[b, a]``) value.
-
-It means, that there's no single type ``Left a`` that makes
-sense without ``Right b``. Only their duality makes sense to us.
-
-In ``python`` we have functions that can be used as type constructors.
-That's why we use ``Success`` and ``Failure`` functions.
-But, when we need to implement
-the behaviour of these types - we use real classes inside.
-That's how we know what to do in each particular case.
-In ``haskell`` we use pattern matching for this.
-
-That's why we have public
-type constructor functions: ``Success`` and ``Failure``
-and internal implementation.
 
 How to compose error types?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/pages/result.rst
+++ b/docs/pages/result.rst
@@ -43,6 +43,7 @@ Make sure to check out how to compose container with
 ``flow`` or :ref:`pipe`!
 Read more about them if you want to compose your containers easily.
 
+
 Pattern Matching
 ----------------
 
@@ -63,13 +64,13 @@ see the example below:
       case Success(10):
           print('Result is "10"')
 
-      # Same as above but using match-by-name
+      # Same as above but using match-by-name  and a different value
       case Success(inner_value=20):
           print('Result is "20"')
 
       # Matches any `Success` instance and binds its value to the ``value`` variable
       case Success(value):
-          print('Result is "{}"'.format(value))
+          print('Result is "{0}"'.format(value))
 
       # Matches if the result stored inside `Failure` is `ZeroDivisionError`
       case Failure(ZeroDivisionError):
@@ -215,6 +216,7 @@ See the example below:
   >>> assert Success(1).map(cast_to_bool) == Success(True)
   >>> assert Success('{"example": "example"}').bind(parse_json) == Success({"example": "example"})
   >>> assert Success('').bind(parse_json).alt(str) == Failure('Expecting value: line 1 column 1 (char 0)')
+
 
 Further reading
 ---------------
--- a/returns/result.py
+++ b/returns/result.py
@@ -2,6 +2,7 @@ from abc import ABCMeta
 from functools import wraps
 from inspect import FrameInfo
 from typing import (
+    TYPE_CHECKING,
     Any,
     Callable,
     ClassVar,
@@ -38,7 +39,7 @@ class Result(
     metaclass=ABCMeta,
 ):
     """
-    Base class for :class:`~_Failure` and :class:`~_Success`.
+    Base class for :class:`~Failure` and :class:`~_Success`.
 
     :class:`~Result` does not have a public constructor.
     Use :func:`~Success` and :func:`~Failure` to construct the needed values.
@@ -50,15 +51,16 @@ class Result(
     """
 
     __slots__ = ('_trace',)
+    __match_args__ = ('_inner_value',)
 
     _inner_value: Union[_ValueType, _ErrorType]
     _trace: Optional[List[FrameInfo]]
 
     # These two are required for projects like `classes`:
     #: Success type that is used to represent the successful computation.
-    success_type: ClassVar[Type['_Success']]
+    success_type: ClassVar[Type['Success']]
     #: Failure type that is used to represent the failed computation.
-    failure_type: ClassVar[Type['_Failure']]
+    failure_type: ClassVar[Type['Failure']]
 
     #: Typesafe equality comparison with other `Result` objects.
     equals = container_equality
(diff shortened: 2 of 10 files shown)


thepabloaguilar:
@sobolevn I've starded working on enabling some Pattern Matching features, the container is `Result` (I'll add support for other containers in this PR) but I'm facing some typing errors 'cause I removed the `Success` and `Failure` functions in order to export the classes!

Could you help me, please?

The first problem is that here: https://github.com/dry-python/returns/pull/981/checks?check_run_id=2867413562

```text
tests/test_result/test_result_map.py:6: error: Call to untyped function "map" in typed context
tests/test_result/test_result_map.py:11: error: Call to untyped function "map" in typed context
tests/test_result/test_result_map.py:12: error: Call to untyped function "alt" in typed context
tests/test_result/test_result_map.py:17: error: Call to untyped function "alt" in typed context
tests/test_result/test_result_bind.py:15: error: Call to untyped function "bind" in typed context
tests/test_result/test_result_bind.py:45: error: Call to untyped function "bind" in typed context
tests/test_result/test_result_bind.py:54: error: Call to untyped function "lash" in typed context
tests/test_result/test_result_bind.py:57: error: Call to untyped function "lash" in typed context
tests/test_result/test_result_bind.py:70: error: Call to untyped function "lash" in typed context
tests/test_functions/test_raise_exception.py:28: error: Call to untyped function "alt" in typed context
tests/test_functions/test_raise_exception.py:33: error: Call to untyped function "alt" in typed context
```
Just using the constructor isn't enough to have a typed context! To fix this I copy-paste the type annotations from `Result` to `Success` and `Failure` and all the mypy tests has passed!

But, when I ran the `typesafety` test more typing errors appeared, the only way I can see to fix all the errors is changing:
- `Success` -> `Result.from_value`
-  `Failure` -> `Result.from_failure`

Because those classmethods works as the old `Success`/`Failure` functions!

docs/pages/result.rst
```diff
@@ -43,6 +43,41 @@ Make sure to check out how to compose container with
...
+      return first_number // second_number
+
+  match div(1, 0):
+      # Matches if the result stored inside `Success` is `10`
+      case Success(10):
+          print('Result is "10"')
+      # Same as above but using match-by-name
+      case Success(inner_value=20):
+          print('Result is "20"')
+      # Matches any `Success` instance and bind its value to the ``value`` variable
```
sobolevn, line 68: `and binds`

docs/pages/result.rst
```diff
@@ -43,6 +43,41 @@ Make sure to check out how to compose container with
...
+
+  @safe
+  def div(first_number: int, second_number: int) -> int:
+      return first_number // second_number
+
+  match div(1, 0):
+      # Matches if the result stored inside `Success` is `10`
+      case Success(10):
+          print('Result is "10"')
+      # Same as above but using match-by-name
```
sobolevn, line 65: Maybe new lines before comments? It might be more readable

docs/pages/result.rst
```diff
@@ -43,6 +43,41 @@ Make sure to check out how to compose container with
...
+      case Success(inner_value=20):
+          print('Result is "20"')
+      # Matches any `Success` instance and bind its value to the ``value`` variable
+      case Success(value):
+          print('Result is "{}"'.format(value))
+      # Matches any `Success` instance
```
sobolevn, line 71: I think that this case can be removed, we already have `catch-all` case above. And having two is an anti-pattern.

sobolevn:
@thepabloaguilar thanks a lot!

Considering your problems:
1. The first one is clear, it happens because our `map` / `lash` / etc functions on `Success` and `Failure` are not annotated. We can use the same annotations we have from `Result`
2. Which leads us to our second problem. It is not quite clear, why this happens. Can you please submit a PR where this is the case?

sobolevn:
Btw, can you please tell me how `sealed` classes work? For example, will this be a typing error?

```python
match div(1, 1):
   case Failure(_):
      print('error')
```

As you can see, we have missed our `Success` case here.

thepabloaguilar:
> Btw, can you please tell me how `sealed` classes work? For example, will this be a typing error?
> 
> ```python
> match div(1, 1):
>    case Failure(_):
>       print('error')
> ```
> 
> As you can see, we have missed our `Success` case here.

Yep, it should raise an exception because you don't stressed all the options! But also I think `mypy` will make this optional, `strict_sealed` 🤔 

Btw, AFIK `sealed` is not implemented yet in `typing`/`typing_extensions`

docs/pages/result.rst
thepabloaguilar: It should be a great `rule` for WPS to check, shouldn't it?

thepabloaguilar:
> @thepabloaguilar thanks a lot!
> 
> Considering your problems:
> 
> 1. The first one is clear, it happens because our `map` / `lash` / etc functions on `Success` and `Failure` are not annotated. We can use the same annotations we have from `Result`
> 2. Which leads us to our second problem. It is not quite clear, why this happens. Can you please submit a PR where this is the case?

@sobolevn I will past the code here 'cause I have more errors to show hahaha

This is the `Success` definition that led to that errors:

```python
@final
class Success(Result[_ValueType, Any]):
    _inner_value: _ValueType

    def __init__(self, inner_value: _ValueType) -> None:
        super().__init__(inner_value)

    def swap(self):
        """Successes swap to :class:`_Failure`."""
        return Failure(self._inner_value)

    def map(
        self,
        function: Callable[[_ValueType], _NewValueType],
    ):
        """Composes current container with a pure function."""
        return Success(function(self._inner_value))

    def apply(
        self,
        container: Kind2[
            'Result',
            Callable[[_ValueType], _NewValueType],
            _ErrorType,
        ],
    ):
        """Calls a wrapped function in a container on this container."""
        if isinstance(container, self.success_type):
            return self.map(container.unwrap())
        return container

    def bind(
        self,
        function: Callable[
            [_ValueType],
            Kind2['Result', _NewValueType, _ErrorType],
        ],
    ):
        """Binds current container to a function that returns container."""
        return function(self._inner_value)

    #: Alias for `bind` method. Part of the `ResultBasedN` interface.
    bind_result = bind

    def lash(
        self,
        function: Callable[
            [_ErrorType], Kind2['Result', _ValueType, _NewErrorType],
        ],
    ):
        """Does nothing for ``Success``."""
        return self

    def alt(
        self,
        function: Callable[[_ErrorType], _NewErrorType],
    ):
        """Does nothing for ``Success``."""
        return self

    def value_or(self, default_value: _NewValueType) -> _ValueType:
        """Returns the value for successful container."""
        return self._inner_value

    def unwrap(self) -> _ValueType:
        """Returns the unwrapped value from successful container."""
        return self._inner_value

    def failure(self) -> NoReturn:
        """Raises an exception for successful container."""
        raise UnwrapFailedError(self)
```

I think that problem is because I haven't copied the return type but when I try to define the return type I got other errors too!

---

Code:
```python
@final
class Success(Result[_ValueType, Any]):
    _inner_value: _ValueType

    def map(
        self,
        function: Callable[[_ValueType], _NewValueType],
    ) -> 'Result[_NewValueType, _ErrorType]':
        """Composes current container with a pure function."""
        return Success(function(self._inner_value))
```

mypy output:
```text
returns/result.py:440: error: Return type "Result[_NewValueType, _ErrorType]" of "map" incompatible with return type "KindN[Success[_ValueType], _UpdatedType, Any, NoReturn]" in supertype "MappableN"
```

And it makes sense (I guess)

---

Code:
```python
@final
class Success(Result[_ValueType, Any]):
    _inner_value: _ValueType

    def map(
        self,
        function: Callable[[_ValueType], _NewValueType],
    ) -> 'Result[_NewValueType, _ErrorType]':
        """Composes current container with a pure function."""
        return Success(function(self._inner_value))
```

mypy output:
```text
returns/result.py:440: error: Return type "Success[_NewValueType]" of "map" incompatible with return type "KindN[Success[_ValueType], _UpdatedType, Any, NoReturn]" in supertype "MappableN"
```

sobolevn:
Try replacing `_ErrorType` with `Any`, because that how you define `Success(Result[_ValueType, Any])`

thepabloaguilar:
Didn't work 😞 

```python
@final
class Success(Result[_ValueType, Any]):
    _inner_value: _ValueType

        def map(
        self,
        function: Callable[[_ValueType], _NewValueType],
    ) -> 'Result[_NewValueType, Any]':
        """Composes current container with a pure function."""
        return Success(function(self._inner_value))
```

```text
returns/result.py:440: error: Return type "Result[_NewValueType, Any]" of "map" incompatible with return type "KindN[Success[_ValueType], _UpdatedType, Any, NoReturn]" in supertype "MappableN"
```

sobolevn:
Ok, I have tried it myself, here's what's wrong:

1. We have two base classes of `Success` where `map` is defined: `MappableN` and` Result`
2. Both of them require different annotations, you cannot make them both happy AFAIU:
```
ex.py:44: error: Return type "..." of "map" incompatible with return type "Result[_NewValueType, Any]" in supertype "Result"
ex.py:44: error: Return type "..." of "map" incompatible with return type "KindN[Success[_ValueType], _UpdatedType, Any, NoReturn]" in supertype "MappableN"
```

3. What `MappableN` expects? `KindN[_MappableType, _UpdatedType, _SecondType, _ThirdType]` which is `KindN[Success, _NewValueType, Any, NoReturn]`
4. What `Result` expects? `Result[_NewValueType]` or `Success[_NewValueType]`
5. It might look like `Success[_NewValueType]` should fit, but no: because we inherit from `SupportsKind2['Result', ...]`, it remembers exactly that tag

My proposal is to be smart 😆  and to ignore all these problems:
- We only need `Success`'s and `Failure`'s annotations, because they have some logic in there. 
- So, we can remove these methods competely or just use something like this:

```python
@final
class Success(Result[_ValueType, Any]):
    def __init__(self, inner_value: _ValueType) -> None:
        ...

    if not TYPE_CHECKING:
        def map(self, function):
            return Success(function(self._inner_value))

x = Success(1)
reveal_type(x)  # Revealed type is "ex.Success[builtins.int*]"
reveal_type(x.map(float))  # Revealed type is "ex.Result[builtins.float*, Any]"
```

What do you think?

thepabloaguilar:
The `if not TYPE_CHECKING:` option seems a little dirty!

I didn't understand well the other option `we can remove these methods completely, could you explain more about it?

sobolevn:
We can put all logic into `Result` itself:

```python
def map(self, function):
    if isinstance(self, Success):
        return function(self._inner_value):
    return self
```

thepabloaguilar:
So, I think I prefer the first solution 😆 

It's just one time `if`

docs/pages/result.rst
```diff
@@ -230,3 +232,4 @@ API Reference
 
 .. automodule:: returns.result
    :members:
+   :exclude-members: Success, Failure
```
sobolevn, line 235: We can still add them, it's ok

returns/result.py
```diff
@@ -309,63 +311,52 @@ def from_result(
...
-        """Composes this container with a function returning container."""
-        return function(self._inner_value)
+        def apply(self, container):
+            """Does nothing for ``Failure``."""
+            return self
 
-    def alt(self, function):
-        """Composes failed container with a pure function to modify failure."""
-        return _Failure(function(self._inner_value))
+        def value_or(self, default_value: _NewValueType) -> _NewValueType:
```
sobolevn, line 353: Let's remove annotation here.

returns/result.py
```diff
@@ -383,63 +374,52 @@ def _get_trace(self) -> Optional[List[FrameInfo]]:
...
+        def apply(self, container):
+            """Calls a wrapped function in a container on this container."""
+            if isinstance(container, self.success_type):
+                return self.map(container.unwrap())
+            return container
 
-    def alt(self, function):
-        """Does nothing for ``Success``."""
-        return self
+        def value_or(self, default_value: _NewValueType) -> _ValueType:
```
sobolevn, line 416: Let's remove annotation here.

sobolevn (commented):
Great work! Questions:
- I think that we should also support `Maybe` and `IOResult`, do you agree?
- Probably we can do it later in other PRs

thepabloaguilar:
> * I think that we should also support `Maybe` and `IOResult`, do you agree?

Yeah, totally! Actually I was thinking in add `Maybe` and `IOResult` in this PR already, but if you think it'll be better to divide in more PRs no problem for me!

thepabloaguilar:
So let's divide!

thepabloaguilar:
Btw, I figured out the reason for Python 3.10 tests be to slow it's related to the `pytest-mypy-plugins`!

But maybe it's related because on Python 3.10 we don't use `nitpick` and some configuration can be missing!

docs/pages/result.rst
```diff
@@ -43,6 +43,42 @@ Make sure to check out how to compose container with
...
+  def div(first_number: int, second_number: int) -> int:
+      return first_number // second_number
+
+  match div(1, 0):
+      # Matches if the result stored inside `Success` is `10`
+      case Success(10):
+          print('Result is "10"')
+
+      # Same as above but using match-by-name
+      case Success(inner_value=20):
```
sobolevn, line 68: Are you sure that this will work? `_inner_value` is protected.

sobolevn: Can we somehow make this a real doctest? And skip it if it's not a `python3.10` environment: https://stackoverflow.com/questions/42708809/python-doctest-skip-a-test-conditionally

thepabloaguilar: > Are you sure that this will work? _inner_value is protected.

Yep, I've tested it! How does it work? I don't know hahaha
Look at here: https://www.python.org/dev/peps/pep-0622/#special-attribute-match-args

```text
...
In deciding what names should be available for matching, the recommended practice is that class patterns should be the mirror of construction; that is, the set of available names and their types should resemble the arguments to __init__().
...
```

sobolevn: What about doctests?

thepabloaguilar: We should do that!! I'm just worried about how to skip, following the link you sent we'll need to put `# doctest: +PYTHON_310` in every single line!

sobolevn: In this case, we can store this file in https://github.com/dry-python/returns/tree/master/tests/test_examples/test_pattern_patching and ignore this folder in `collect` hook of `pytest` https://docs.pytest.org/en/6.2.x/reference.html#collection-hooks

thepabloaguilar: I'll do it tonight!

tests/conftest.py
```diff
@@ -1,2 +1,38 @@
...
+        Path('tests/test_examples/test_result/test_pattern_matching.py'),
+    ],
+    (3, 8): [
+        Path('tests/test_examples/test_result/test_pattern_matching.py'),
+    ],
+    (3, 9): [
+        Path('tests/test_examples/test_result/test_pattern_matching.py'),
+    ],
+}
+paths_to_ignore_now = [
```
sobolevn, line 23: `frozenset`?

tests/conftest.py
```diff
@@ -1,2 +1,38 @@
...
+from typing import Optional
+
+from _pytest.config import Config
+from py import path as pypath
+
 # See https://github.com/HypothesisWorks/hypothesis/pull/2567
 pytest_plugins = ['hypothesis.extra.pytestplugin']
+
+python_version = (sys.version_info.major, sys.version_info.minor)
+paths_to_ignore = {  # noqa: WPS407
```
sobolevn, line 12: `types.MutableMapping`?

tests/conftest.py
```diff
@@ -1,2 +1,38 @@
...
+from pathlib import Path
+from typing import Optional
+
+from _pytest.config import Config
+from py import path as pypath
+
 # See https://github.com/HypothesisWorks/hypothesis/pull/2567
 pytest_plugins = ['hypothesis.extra.pytestplugin']
+
+python_version = (sys.version_info.major, sys.version_info.minor)
```
sobolevn, line 11: Please, don't forget to type all these vars as `Final`

tests/conftest.py
```diff
@@ -1,2 +1,38 @@
...
+paths_to_ignore = {  # noqa: WPS407
+    (3, 7): [
+        Path('tests/test_examples/test_result/test_pattern_matching.py'),
+    ],
+    (3, 8): [
+        Path('tests/test_examples/test_result/test_pattern_matching.py'),
+    ],
+    (3, 9): [
```
sobolevn, line 19: maybe `frozenset` instead of a `list`?

sobolevn (changes requested):
Great work! Only some minor polishing is left! 👍

tests/conftest.py
thepabloaguilar: It's `Final` too, isn't it?

thepabloaguilar: Oh, it's `types` no `typing`!

Maybe [types.MappingProxyType](https://docs.python.org/3/library/types.html#types.MappingProxyType) should be better!

sobolevn (approved):
Great work, @thepabloaguilar!

PR #990, opened 2021-06-28
Enables Pattern Matching support for `IOResult` container

thepabloaguilar:
# Enables Pattern Matching support for `IOResult` container

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ See [0Ver](https://0ver.org/).
 
 - Enables Pattern Matching support for `Result` containers
 - Enables Pattern Matching support for `Maybe` container
+- Enables Pattern Matching support for `IOResult` container
 
 ## 0.16.0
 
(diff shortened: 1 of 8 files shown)


sobolevn (changes requested):
Great, thanks!

Let's write several unit tests for all three types. THey obviously should be skipped for `python<3.10`

thepabloaguilar:
Which type of test? I didn't understand the goal

sobolevn:
Something like this:

```python
def test_pattern_matching_result():
    match some_func(1):
        case Success(inner_value=1):
             assert inner_value == 1
    pytest.fail('Was not matched')
```

And the same for all 3 monads and each of their subclasses. I want to test that `__match_args__` work basically 👍

thepabloaguilar:
What do you think to add those tests in another PR? It seems more clear to review later

sobolevn:
Sure! 👍

sobolevn (approved):

sobolevn:
Thanks!

PR #991, opened 2021-07-01
Adds Pattern Matching tests & fixes pattern matching examples

thepabloaguilar:
# Adds Pattern Matching tests & fixes pattern matching examples

## Checklist

### DIFF
--- a/tests/test_examples/test_io/test_ioresult_container/test_ioresult_pattern_matching.py
+++ b/tests/test_examples/test_io/test_ioresult_container/test_ioresult_pattern_matching.py
@@ -1,9 +1,13 @@
 from returns.io import IOFailure, IOResult, IOSuccess
+from returns.result import Success
 
 container: IOResult[int, str] = IOSuccess(42)
 match container:
     # Matches if the result stored inside `IOSuccess` is `42`
-    case IOSuccess(42):
+    # We need to use `Success` until the custom matching protocol
+    # is released. For more information, please visit:
+    # https://www.python.org/dev/peps/pep-0622/#custom-matching-protocol
+    case IOSuccess(Success(42)):
         print('Result is "42"')
 
     # Matches any `IOSuccess` instance
--- a/tests/test_examples/test_maybe/test_maybe_pattern_matching.py
+++ b/tests/test_examples/test_maybe/test_maybe_pattern_matching.py
@@ -37,5 +37,5 @@ if __name__ == '__main__':
             print('Book found: {0}'.format(book.name))
 
         # Matches `Nothing` instance
-        case Nothing:
+        case Maybe.empty:
             print('Not found the desired book!')
--- a/tests/test_pattern_matching.py
+++ b/tests/test_pattern_matching.py
@@ -1,0 +1,73 @@
+import pytest
+
+from returns.io import IO, IOFailure, IOResult, IOSuccess
+from returns.maybe import Maybe, Nothing, Some
+from returns.result import Failure, Result, Success
+
+
+@pytest.mark.parametrize('container', [
+    Success(10),
+    Success(42),
+    Failure(RuntimeError()),
+    Failure(Exception()),
+])
+def test_result_pattern_matching(container: Result[int, Exception]):
+    """Ensures ``Result`` containers work properly with pattern matching."""
+    match container:
+        case Success(10):
+            assert isinstance(container, Success)
+            assert container.unwrap() == 10
+        case Success(value):
+            assert isinstance(container, Success)
+            assert value == 42
+            assert container.unwrap() == value
+        case Failure(RuntimeError()):
+            assert isinstance(container, Failure)
+            assert isinstance(container.failure(), RuntimeError)
+        case Failure(_):
+            assert isinstance(container, Failure)
+            assert isinstance(container.failure(), Exception)
+        case _:
+            pytest.fail('Was not matched')
+
+
+@pytest.mark.parametrize('container', [
+    Some('SOME'),
+    Some('THERE IS SOME VALUE'),
+    Nothing,
+])
+def test_maybe_pattern_matching(container: Maybe[str]):
+    """Ensures ``Maybe`` containers work properly with pattern matching."""
+    match container:
+        case Some('SOME'):
+            assert isinstance(container, Some)
+            assert container.unwrap() == 'SOME'
+        case Some(value):
+            assert isinstance(container, Some)
+            assert value == 'THERE IS SOME VALUE'
+            assert container.unwrap() == value
+        case Maybe.empty:
+            assert container is Nothing
+        case _:
+            pytest.fail('Was not matched')
+
+
+@pytest.mark.parametrize('container', [
+    IOSuccess(42.0),
+    IOSuccess(10.0),
+    IOFailure(50),
+])
+def test_ioresult_pattern_matching(container: IOResult[float, int]):
+    """Ensures ``IOResult`` containers work properly with pattern matching."""
+    match container:
+        case IOSuccess(Success(42.0)):
+            assert isinstance(container, IOSuccess)
+            assert container.unwrap() == IO(42.0)
+        case IOSuccess(value):
+            assert isinstance(container, IOSuccess)
+            assert container.unwrap() == IO(value.unwrap())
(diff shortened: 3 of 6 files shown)


tests/test_examples/test_io/test_ioresult_container/test_ioresult_pattern_matching.py
```diff
@@ -1,9 +1,13 @@
...
+from returns.result import Success
 
 container: IOResult[int, str] = IOSuccess(42)
 match container:
     # Matches if the result stored inside `IOSuccess` is `42`
-    case IOSuccess(42):
+    # We need to use `Success` until the custom matching protocol
+    # is released. For more information, please visit:
+    # https://www.python.org/dev/peps/pep-0622/#custom-matching-protocol
+    case IOSuccess(Success(42)):
```
thepabloaguilar, line 10: @sobolevn I think we can make a workarround to make possible to do:
```python
match container:
    case IOSuccess(42):
        ...
```

After each `bind`, `map`, and etc we save the value in another variable than `_inner_value` and put it in `__match_args__` instead of `_inner_value`!
But also, we can wait for the __Custom Matching Protocol__!

thepabloaguilar:
@sobolevn You were right, this:
```python
match container:
    Success(inner_value=42):
        ...
```

Does not work, that example will fall on `Success(value)`, but this works:

```python
match container:
    Success(_inner_value=42):
        ...
```

So, I've removed this example. I don't think that's a good practice to use an inner attribute and also, users don't need it because they can use only positional matching since there's only one position!

tests/test_pattern_matching.py
```diff
@@ -0,0 +1,73 @@
+import pytest
+
+from returns.io import IO, IOFailure, IOResult, IOSuccess
+from returns.maybe import Maybe, Nothing, Some
+from returns.result import Failure, Result, Success
+
+
+@pytest.mark.parametrize('container', [
```
sobolevn, line 8: Let's use `skipif` here, it would be more readable! 👍

tests/test_examples/test_maybe/test_maybe_pattern_matching.py
```diff
@@ -37,5 +37,5 @@ def _find_book(book_id: int) -> Maybe[_Book]:
             print('Book found: {0}'.format(book.name))
 
         # Matches `Nothing` instance
-        case Nothing:
+        case Maybe.empty:
```
sobolevn, line 40: Why not direct `Nothing`? 🤔

tests/test_pattern_matching.py
```diff
@@ -0,0 +1,73 @@
...
+    """Ensures ``Maybe`` containers work properly with pattern matching."""
+    match container:
+        case Some('SOME'):
+            assert isinstance(container, Some)
+            assert container.unwrap() == 'SOME'
+        case Some(value):
+            assert isinstance(container, Some)
+            assert value == 'THERE IS SOME VALUE'
+            assert container.unwrap() == value
+        case Maybe.empty:
```
sobolevn, line 49: The same about `.empty`

tests/test_pattern_matching.py
```diff
@@ -0,0 +1,73 @@
...
+
+@pytest.mark.parametrize('container', [
+    IOSuccess(42.0),
+    IOSuccess(10.0),
+    IOFailure(50),
+])
+def test_ioresult_pattern_matching(container: IOResult[float, int]):
+    """Ensures ``IOResult`` containers work properly with pattern matching."""
+    match container:
+        case IOSuccess(Success(42.0)):
```
sobolevn, line 63: Is there a way to match this as `IOSuccess(42)`?

sobolevn (changes requested):
Thanks! Great work!

tests/test_pattern_matching.py
sobolevn: Instead on custom `pytest` hook.

tests/test_examples/test_maybe/test_maybe_pattern_matching.py
thepabloaguilar: I think this can be an bug, I'll report later to Python team

But, I've changed because we can only use Types here and `Nothing` it's an object. BUT `Maybe.empty` is working and it's a `Nothing` instance under the hoods! So I think it's a bug

tests/test_pattern_matching.py
thepabloaguilar: Yep, I've wrote about this in a previous comment

sobolevn: Oh, I see: https://github.com/dry-python/returns/pull/991#discussion_r661976248

Ok, let's keep it this way for now!

thepabloaguilar: Ok

tests/test_examples/test_maybe/test_maybe_pattern_matching.py
thepabloaguilar: Hummm... Maybe I know what's happening, look: [Constant Value Patterns](https://www.python.org/dev/peps/pep-0622/#constant-value-patterns)

thepabloaguilar: Anyway, I've opened an issue there: https://bugs.python.org/issue44617

sobolevn: Try `.Nothing`?

thepabloaguilar: Does not work! 😢

thepabloaguilar: Ok, this is why `Maybe.empty` works but `Nothing` doesn't:
https://bugs.python.org/issue44617#msg397459

tests/test_pattern_matching.py
thepabloaguilar: I think I just need to fix this one here, I'll try to do it tomorrow

tests/test_examples/test_maybe/test_maybe_pattern_matching.py
sobolevn: > So your `case Nothing` example is an irrefutable capture pattern that binds the subject to the name "Nothing", while `case Maybe.empty` is a value pattern.

Pattern matching syntax 🤦

tests/test_pattern_matching.py
thepabloaguilar: I've add a marker `min_python_3_10`, I think it's more readable than `skipif`

thepabloaguilar: I've added a marker `min_python_3_10`, I think it's more readable than `skipif`

thepabloaguilar: Well, it doesn't work. `skipif` neither

thepabloaguilar: Python evaluates the function body

tests/test_examples/test_maybe/test_maybe_pattern_matching.py
thepabloaguilar: I got a `SyntaxError: invalid syntax`

tests/test_pattern_matching.py
thepabloaguilar: As you can see here: https://github.com/dry-python/returns/pull/991/checks?check_run_id=3563951538

We had a problem in the collect part

thepabloaguilar: I've rolled back!

tests/test_examples/test_maybe/test_maybe_pattern_matching.py
thepabloaguilar: Looking at the reference I think it is not possible to do that than `Maybe.empty`:
https://www.python.org/dev/peps/pep-0634/#appendix-a-full-grammar

sobolevn: Got it 😞 
https://www.python.org/dev/peps/pep-0636/#matching-specific-values

sobolevn (approved):
Awesome work! 

Feel free to merge when you need it.

PR #1096, opened 2021-10-06
Update to python 3.10 + clean up in test-workflow

### LINKED ISSUE #1095: Switch `3.10-rc2` to `3.10` in GitHub Actions
We need to:
1. delete these lines https://github.com/dry-python/returns/blob/master/.github/workflows/test.yml#L21-L24
2. remove `continue-on-error`: https://github.com/dry-python/returns/blob/master/.github/workflows/test.yml#L17
3. add `3.10` here https://github.com/dry-python/returns/blob/master/.github/workflows/test.yml#L20

jonbrohauge:
close #1095

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, 3.10]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
 
     steps:
     - uses: actions/checkout@v2


.github/workflows/test.yml
```diff
@@ -14,14 +14,9 @@ on:
...
     runs-on: ubuntu-latest
-    continue-on-error: ${{ matrix.allow-failure }}
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
-        allow-failure: [false]
-        include:
-          - python-version: 3.10.0-rc.2
-            allow-failure: true
+        python-version: [3.7, 3.8, 3.9, 3.10]
```
sobolevn, line 19: It should be `['3.7', '3.8', '3.9', '3.10']`

sobolevn:
Thanks a lot!

Issue #1114, opened 2021-11-02
Adding some utility methods to Maybe/Result


ariebovenberg:
Have you considered adding some handy utility functions to `Maybe` and `Result`?
In using `returns` I'm missing many helpful methods you can find in Rust, for example.
Adding a few of these could increase the usability of the containers of `returns`.

Related: #1091

Here they are, in my subjective order of usefulness:

## Setdefault

Set a value if one is not yet present. 
The name `setdefault` would be consistent with the python dict method.
(Although it might be good to choose another name to make clear that
the method doesn't mutate.)

```python
>>> Some(6).setdefault(7)
Some(6)
>>> Nothing.setdefault(3)
Some(3)
```

## Zip

Combine the values from two containers into a tuple, if both are successful.

```python
>>> Some(8).zip(Some(2))
Some((8, 2))
>>> Some(7).zip(Nothing)
Nothing
>>> Nothing.zip(Some(3))
Nothing
>>> Nothing.zip(Nothing)
Nothing
```

## Filter

Keep a value based on a predicate

```python
>>> Some("9").filter(str.isdigit)
Some("9")
>>> Some("foo").filter(str.isdigit)
Nothing
>>> Nothing.filter(str.isdigit)
Nothing
```

## And

`AND` logic operation

```python
>>> Some(5).and_(Some(9))
Some(9)
>>> Some(9).and_(Nothing)
Nothing
>>> Nothing.and_(Some(8))
Nothing
```

## Or 

`OR` logic operation

```python
>>> Some(5).or_(Some(9))
Some(5)
>>> Some(9).or_(Nothing)
Some(9)
>>> Nothing.or_(Some(8))
Some(8)
>>> Nothing.or_(Nothing)
Nothing
```

CucumisSativus:
I think `Setdefault` is already in the library, in the form of `value_or()`.
With zip (or map_n) and filter I really think those would be valuable additions to the library.
Could you show how would you use `and` and `or`, I think I am missing the idea how could those be used in the real source code

ariebovenberg:
I'll illustrate with this example:

```python
from typing import NamedTuple


class User(NamedTuple):
    id: int
    nickname: Maybe[str]
    first_name: Maybe[str]
    last_name: str
    joined: datetime
    job: Maybe[str]
    location: Maybe[int]
```

## Setdefault

[...]

PR #1131, opened 2021-11-16
Issue 1114

CucumisSativus:
# I have made things!

## Checklist

## Related issues

Refs #1114 

A simple implementation of filter for maybe. I am happy to implement something similar for Result, but I'm not entirely sure about the preferred approach. 
In Scala it looks more or less like this 
```
      if (p(value)) this else Failure(new NoSuchElementException("Predicate does not hold for " + value))
```
Not sure if something like this is ok for Python as well. 

My second idea was to have a signature like
```
def filter(predicate, on_failure):
(...)
```
so user can specify exactly what happens in case of not matching predicate


sobolevn:
We also need to make an interface type for `filter`. This way we will be able to use with HKT.
See `interfaces/` folder.

CucumisSativus:
@sobolevn I think i managed to add them, but lets say i'm not 100% sure about those. Looking forward for some feedback :)

returns/interfaces/filterable.py
```diff
@@ -0,0 +1,41 @@
...
+        ...     return argument % 2 == 0
+
+        >>> assert filter_(example)(Some(5)) == Nothing
+        >>> assert filter_(example)(Some(6)) == Some(6)
+        >>> assert filter_(example)(Nothing) == Nothing
+
+    """
+
+    @abstractmethod
+    def filter(
```
sobolevn, line 37: Future note: we can also make an `overload` for `TypeGuard` case.
See https://github.com/python/typeshed/pull/6140

returns/maybe.py
```diff
@@ -430,10 +436,10 @@ def failure(self):
         raise UnwrapFailedError(self)
 
     def filter(self, function):
```
sobolevn, line 438: Is it possible to create a common `filter_` function for all possible types? `Result`, `IOResult`, ...?

sobolevn (commented):
Thanks a lot! Just one question 🙂

returns/maybe.py
ariebovenberg: `filter` is special in the sense that it 'empties' the container, which is possible in `Maybe`, but not in `Result` (i.e. what would the failure value of `Success(4).filter(lambda x: False)` be?)

sobolevn: Ok, I see! Than, it looks like we can use `MaybeLikeN` to limit the input type of `filter_`?

returns/interfaces/filterable.py
CucumisSativus: Could you elaborate a bit what can we do with type guard here? We are not filtering for example a list with different types. Filter executed on `_FilterableType` always returns `_FilterableType`. Did I miss something?

returns/maybe.py
CucumisSativus: @sobolevn I used it, please tell me if its something you were looking for. I am not entirely sure if I did it right :)

returns/interfaces/filterable.py
sobolevn: Nevermind, this was a self note 🙂 
This is even not supported by mypy yet.

CucumisSativus:
@sobolevn when i'm running `poetry run pytest typesafety -p no:cov -o addopts="" --mypy-ini-file=setup.cfg` locally I get `830 passed, 1 warning`. Any ideas why does it fail on github?

sobolevn:
I will take a look a bit later this week.

Thanks a lot for the PR!

returns/interfaces/filterable.py
```diff
@@ -0,0 +1,54 @@
...
+        self: _FilterableType,
+        predicate: Callable[[_FirstType], bool],
+    ) -> Kind1[_FilterableType, _FirstType]:
+        """Applies 'predicate' to the result fo a previous computation."""
```
ariebovenberg, line 44: typo: fo->of

returns/interfaces/filterable.py
```diff
@@ -0,0 +1,54 @@
...
+    Not all types can be ``Filterable`` because we require
+    a possibility to access internal value and to model a case,
+    where the predicate is false
+
+    .. code:: python
+
+        >>> from returns.maybe import Nothing, Some
+        >>> from returns.pointfree import filter_
+
+        >>> def example(argument: int) -> bool:
```
ariebovenberg, line 30: what would you think of naming the example function? Naming the function `is_even` makes the example clearer in my mind

CucumisSativus:
@sobolevn i think I found it

djbrown:
any news here? I would love to use those utilities 🤩

returns/interfaces/filterable.py
ariebovenberg: It looks like mypy [supports](https://mypy.readthedocs.io/en/latest/type_narrowing.html#user-defined-type-guards) `TypeGuard` now 🎉 :

PR #1147, opened 2021-11-27
add missing __slots__ to containers and their base classes (#1144)

### LINKED ISSUE #1144: `__slots__ = ()` missing in many base classes?
related: #416

It seems an empty `__slots__` is forgotten in most base classes (like `Immutable`, `Equable`, etc.), greatly reducing the advantage (i.e. memory savings) of defining `__slots__` in their subclasses (e.g. `Result`, `Maybe`). Unless I'm missing something, this is not intended. In that case I will gladly submit a PR fixing this.

# Example

```python
>>> from returns.result import Success  # Result defines __slots__
>>> a = Success(4)
>>> a.__dict__  # this should not be present, but it is inherited anyway from base classes without __slots__
{}
>>> object.__setattr__(a, 'foo', 9)  # We can technically set any attribute because there is a `__dict__`
>>> a.foo
9
>>> a.__dict__
{'foo': 9}
```

# Expected behavior of slots classes

```python
>>> class Base:
...  __slots__ = ()  # necessary!
>>> class A(Base):
...  __slots__ = ('foo', )
...
>>> a = A()
>>> a.foo = 4
>>> a.bar = 9
  File "<stdin>", line 1, in <module>
AttributeError: 'A' object has no attribute 'bar'
>>> a.__dict__
  File "<stdin>", line 1, in <module>
AttributeError: 'A' object has no attribute '__dict__'
```

ariebovenberg:
# I have made things!

- slots added to many _many_ classes
- in pytest plugin, the mechanism for tracking 'errors handled' needed a rework
- test added to prevent forgetting slots in the future

## Checklist

## Related issues

- Closes #1144

## Open questions

In adding `__slots__`, internal changes were needed to the `pytest` plugin.  It works now (as discussed in #1144), but I still have a nagging feeling (monkey-patching and mutable global state will do that... 😉 )

I would like to suggest a follow-up change to the pytest plugin: that it undoes its patches after each test (like `unittest.mock.patch` does). The advantages are:
- classes are only patched when using the special `ReturnsAsserts`. This means it 100% doesn't affect any other tests (i.e. we get more predictable behavior). Currently all tests running _after_ a `returns`-enabled test are patched, those before are not. A very nasty thing to debug if it ever messes with anybody's tests 👀 
- the 'errors handled' mapping doesn't need to be a mutable global variable (👿) , it can be stored in the `ReturnsAsserts` instance.
- the 'errors handled' mapping doesn't need to be cleared after each test 🚀

🤔 Of course, we'll have to see what the performance impact is. Perhaps we might need to split the performance-heavy patching into a separate fixture, so it's only activated when needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Versions before `1.0.0` are `0Ver`-based:
 incremental in minor, bugfixes only are patches.
 See [0Ver](https://0ver.org/).
 
+## 0.17.1
+
+### Bugfixes
+
+- Fixes `__slots__` not being set properly in containers and their base classes
 
 ## 0.17.0
 
--- a/returns/context/requires_context.py
+++ b/returns/context/requires_context.py
@@ -77,6 +77,8 @@ class RequiresContext(
 
     """
 
+    __slots__ = ()
+
     #: This field has an extra 'RequiresContext' just because `mypy` needs it.
     _inner_value: Callable[[RequiresContext, _EnvType], _ReturnType]
 
--- a/returns/context/requires_context_future_result.py
+++ b/returns/context/requires_context_future_result.py
@@ -95,6 +95,8 @@ class RequiresContextFutureResult(
 
     """
 
+    __slots__ = ()
+
     #: Inner value of `RequiresContext`
     #: is just a function that returns `FutureResult`.
     #: This field has an extra 'RequiresContext' just because `mypy` needs it.
--- a/returns/context/requires_context_ioresult.py
+++ b/returns/context/requires_context_ioresult.py
@@ -102,6 +102,8 @@ class RequiresContextIOResult(
 
     """
 
+    __slots__ = ()
+
     #: Inner value of `RequiresContext`
     #: is just a function that returns `IOResult`.
     #: This field has an extra 'RequiresContext' just because `mypy` needs it.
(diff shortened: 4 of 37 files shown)


sobolevn (approved):
LGTM! Thanks!

sobolevn:
> classes are only patched when using the special ReturnsAsserts. This means it 100% doesn't affect any other tests (i.e. we get more predictable behavior). Currently all tests running after a returns-enabled test are patched, those before are not. A very nasty thing to debug if it ever messes with anybody's tests 

Agreed!

> the 'errors handled' mapping doesn't need to be a mutable global variable (👿) , it can be stored in the ReturnsAsserts instance.

Yes, I thought so too during code review.

ariebovenberg:
It seems that for different python versions, `flake8` attributes `WPS602` errors to different lines of code. Double `noqa`s were needed in `pytest` plugin. @sobolevn you can probably tell if this is a problem with the style guide, or flake8?

```python
# passes in py37, but not py38+
class A:
    @staticmethod  # noqa: WPS602
    def foo():
        ...

# passes in py38+, but not py37
class A:
    @staticmethod
    def foo():    # noqa: WPS602
        ...

# passes on all versions:
class A:
    @staticmethod  # noqa: WPS602
    def foo():  # noqa: WPS602
        ...
```

sobolevn:
@ariebovenberg yes, this is true 😞

sobolevn (approved):
I agree with your WPS issue, that things like `__bases__`, `__module__`, etc should be allowed.

Thanks a lot for working on this! 🎉

PR #1164, opened 2021-12-07
cleanup patches after test in pytest plugin (#1148)

### LINKED ISSUE #1148: Improved monkeypatching in pytest plugin
The pytest plugin patches containers to track errors handled. As described in #1147, there are two problems with the current approach:
- the monkeypatch is not undone after test execution
- a global state is used to track errors handled.

These can be solved by:
1. undoing the monkeypatch after each test where it is used (perhaps using `unittest.mock.patch`)
2. storing the errors handled in the fixture instance (only possible after (1))

This proposal has already been approved. I will create a PR soon.

ariebovenberg:
# I have made things!

As discussed in #1148 and #1147 

excuse the large diff; I took the liberty of rearranging the functions in the module to:
- group them together by functionality
- remove the need for a class which only had staticmethods

## Checklist

## Related issues

Closes #1148

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -20,3 +20,7 @@ hypothesis==6.30.0
 # TODO: Remove this lock when we found and fix the route case.
 # See: https://github.com/typlog/sphinx-typlog-theme/issues/22
 jinja2==3.0.3
+
+# TODO: Remove this lock when this dependency issue is resolved.
+# See: https://github.com/miyakogi/m2r/issues/66
+mistune<2.0.0
--- a/returns/contrib/pytest/plugin.py
+++ b/returns/contrib/pytest/plugin.py
@@ -1,23 +1,25 @@
 import inspect
 import sys
-from contextlib import contextmanager
+from contextlib import ExitStack, contextmanager
 from functools import partial, wraps
 from types import FrameType
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterator, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Generator,
+    Iterator,
+    TypeVar,
+    Union,
+)
+from unittest import mock
 
 import pytest
 from typing_extensions import Final, final
 
 if TYPE_CHECKING:
     from returns.interfaces.specific.result import ResultLikeN
-
-_ERROR_HANDLERS: Final = (
-    'lash',
-)
-_ERRORS_COPIERS: Final = (
-    'map',
-    'alt',
-)
 
 # We keep track of errors handled by keeping a mapping of <object id>: object.
 # If an error is handled, it is in the mapping.
@@ -28,7 +30,7 @@ _ERRORS_COPIERS: Final = (
 # Also, the object itself cannot be (in) the key because
 # (1) we cannot always assume hashability and
 # (2) we need to track the object identity, not its value
-_ERRORS_HANDLED: Final[Dict[int, Any]] = {}  # noqa: WPS407
+_ErrorsHandled = Dict[int, Any]
 
 _FunctionType = TypeVar('_FunctionType', bound=Callable)
 _ReturnsResultType = TypeVar(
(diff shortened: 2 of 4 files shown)


ariebovenberg:
Apparently there is an issue in the `m2r` dependency chain. Reports are coming in, see https://github.com/miyakogi/m2r/issues/66.

For now pinned the offending dependency.

returns/contrib/pytest/plugin.py
```diff
@@ -28,7 +30,7 @@
 # Also, the object itself cannot be (in) the key because
 # (1) we cannot always assume hashability and
 # (2) we need to track the object identity, not its value
-_ERRORS_HANDLED: Final[Dict[int, Any]] = {}  # noqa: WPS407
+_ErrorsHandled = Dict[int, Any]
```
sobolevn, line 33: Oh, this is now a type alias. Let's leave a `TODO` item to use explicit `TypeAlias` once `mypy@0.920` is out.

returns/contrib/pytest/plugin.py
```diff
@@ -166,65 +118,107 @@ def _trace_function(
...
+
+@pytest.fixture()
+def returns() -> Generator[ReturnsAsserts, None, None]:
+    """Returns class with helpers assertions to check containers."""
+    with _spy_error_handling() as errors_handled:
+        yield ReturnsAsserts(errors_handled)
+
+
+@contextmanager
+def _spy_error_handling() -> Generator[_ErrorsHandled, None, None]:
```
sobolevn, line 149: Looks like it is just `Iterator[_ErrorsHandled]`

returns/contrib/pytest/plugin.py
```diff
@@ -166,65 +118,107 @@ def _trace_function(
...
+    else:
+        def wrapper(self, *args, **kwargs):
+            original_result = original(self, *args, **kwargs)
+            if id(self) in errs:
+                errs[id(original_result)] = original_result
+            return original_result
+    return wraps(original)(wrapper)  # type: ignore
+
+
+_ERROR_HANDLING_PATCHERS: Final = {  # noqa: WPS407
```
sobolevn, line 220: Do we need this to be mutable? Or will `types.MappingProxyType` work here?

sobolevn (commented):
Great! Thank you!

A couple of questions from me 🙂

returns/contrib/pytest/plugin.py
ariebovenberg: I was doubting as well whether to use it. Based on the PEP it seemed optional -- necessary in case of forward references, for example.

On the other hand, if this is looking to be the default way of declaring type aliases (explicit!) then we can already import from `typing_extensions` right?

returns/contrib/pytest/plugin.py
ariebovenberg: Indeed -- I assumed `contextmanager` would complain -- let's test that hypothesis 🧑‍🔬 ...

returns/contrib/pytest/plugin.py
ariebovenberg: Nope. Chose the `dict` for simplicity. Can use a readonly proxy just fine

returns/contrib/pytest/plugin.py
ariebovenberg: Hmmm something strange in `mypy` indeed, it gives:

```
Variable "typing_extensions.TypeAlias" is not valid as a type
```

mypy doesn't support them yet at all yet?

sobolevn: Yes, `0.910` does not support it yet, `0.920` will (hopefully).

returns/contrib/pytest/plugin.py
sobolevn: Let's do this! We are kinda strict on mutability in this specific project 👍

ariebovenberg: Naturally 😅 . It's a shame`MappingProxyType` isn't a full `frozendict`-like object, but we'll take it.

returns/contrib/pytest/plugin.py
ariebovenberg: Is there a other features coming in 0.920 that returns finds useful?

docs/requirements.txt
```diff
@@ -20,3 +20,7 @@ hypothesis==6.30.0
 # TODO: Remove this lock when we found and fix the route case.
 # See: https://github.com/typlog/sphinx-typlog-theme/issues/22
 jinja2==3.0.3
+
+# TODO: Remove this lock when this dependency issue is resolved.
```
sobolevn, line 24: Extra thanks for this! 👍

returns/contrib/pytest/plugin.py
sobolevn: I am not sure about `ParamSpec` (it might not be released just yet), but it would remove `decorator` plugin! 🎉

sobolevn (approved):
Great work!

PR #1217, opened 2022-01-07
Adds `attempt` decorator

### LINKED ISSUE #264: Variant of safe that preserves the argument that lead to an exception
I propose a function/decorator (maybe `attempt` would be a good name?) similar to `safe` with the twist of returning `Failure(bad_input)` from `safe(f)(bad_input)` instead of `Failure(SomeException)`. An example where this would be useful:

```python
(
    attempt(json.loads)('<hello world>')
    .alt(lambda s: ValueError(f'"{s}" is not a valid json')) 
)
# Failure(ValueError('"<hello world>" is not a valid json'))
```

thepabloaguilar:
# Adds `attempt` decorator

## Checklist

## Related issues

Closes #264


docs/pages/result.rst
```diff
@@ -116,6 +116,28 @@ If you want to `safe` handle only a set of exceptions:
     ...
   ValueError: Too big
 
+attempt
+~~~~~~~
+
+Similar to :func:`safe <returns.result.safe>` function but instead
+of wrapping the exception error in a Failure container it'll wrap the
```
sobolevn, line 123:
```suggestion
of wrapping the exception error in a :func:`~Failure` container it'll wrap the
```

typesafety/test_result/test_attempt.yml
```diff
@@ -0,0 +1,34 @@
...
+- case: safe_decorator_one_param
+  disable_cache: false
+  main: |
+    from returns.result import attempt
+
+    @attempt
+    def test(param: str) -> int:
+        return 1
+
+    reveal_type(test)  # N: Revealed type is "def (builtins.str*) -> returns.result.Result[builtins.int*, builtins.str*]"
```
sobolevn, line 22: Let's also add this example:

```python
def other(param: int) -> str:
    return 'a'

reveal_type(attempt(other))
```

sobolevn (commented):
Thanks you! Only minor suggestions.

I've also removed a test file, because doctests are good enough in this case.

docs/pages/result.rst
thepabloaguilar: I've added a class reference since the exposed `Failure` is a class:
```text
:class:`Failure <returns.result.Failure>`
```

sobolevn (approved):
Great work! Thank you!

PR #1251, opened 2022-01-24
Update test.yml

sobolevn:


sobolevn:
It works! https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value

PR #1360, opened 2022-04-13
Use format string in pattern matching example

johnthagen:
## Checklist

## Summary

`test_result_pattern_matching.py` shows up in the generated docs. Simplify string formatting by using format strings.

- https://returns.readthedocs.io/en/latest/pages/result.html#pattern-matching


sobolevn (changes requested):
Sorry, but we use https://github.com/wemake-services/wemake-python-styleguide that does not allow `f`-strings.

PR #1469, opened 2022-08-16
Fixes custom pickle protocol to handle `None` values as well

### LINKED ISSUE #1405: Success(None) and Some(None) fails to load from pickle
# Bug report

## What's wrong

It appears returns containers fail to load from serialized pickle form when the inner value is `None`. 

I wrote a quick test script:
```python
import pickle

from returns.result import Success, Failure
from returns.maybe import Some

assert(pickle.loads(pickle.dumps(Success(None))) == Success(None))
assert(pickle.loads(pickle.dumps(Failure(None))) == Failure(None))
assert(pickle.loads(pickle.dumps(Some(None)))    == Some(None))
```

Here's the stacktrace
```console
$ python ./test.py
  File "/home/yuki/cum/./test.py", line 6, in <module>
    assert(pickle.loads(pickle.dumps(Success(None))) == Success(None))
  File "/nix/store/rsyzznrv5jg9dx7yj26lf4yjybmafba2-python3-3.9.12-env/lib/python3.9/site-packages/returns/primitives/container.py", line 34, in __eq__
    return container_equality(self, other)  # type: ignore
  File "/nix/store/rsyzznrv5jg9dx7yj26lf4yjybmafba2-python3-3.9.12-env/lib/python3.9/site-packages/returns/primitives/container.py", line 61, in container_equality
    self._inner_value == other._inner_value,  # type: ignore # noqa: WPS437
AttributeError: _inner_value
```

So far I've tested `Success`, `Failure`, `Some` and all three of them fail the same way.

## How is that should be

As shown in the test script, returns container should be "recreated" properly including its `None` inner value from the pickled form.

## System information

- `python` version: `3.9.12`
- `returns` version: `0.19.0`
- `mypy` version: `0.950`

thepabloaguilar:
# Fixes custom pickle protocol to handle `None` values as well

I've created an intermediate representation to Pickle/Unpickle objects to handle `None` values too! What's happening now is:
* Pickle calls `__getstate__`, which returns a value (that can be `None`)
* Pickle calls `__setstate__` __ONLY__ the value is not `None` in the `__getstate__` phase

```text
Note If __getstate__() returns a false value, the __setstate__() method will not be called upon unpickling.
```
[source](https://docs.python.org/3/library/pickle.html#object.__setstate__)

With the changes introduced by this PR all values are wrapped in a dict, this way we preserve the value even if it's `None`!

## Checklist

## Related issues

Closes #1405

### DIFF
--- a/tests/test_primitives/test_container/test_base_container/test_pickle.py
+++ b/tests/test_primitives/test_container/test_base_container/test_pickle.py
@@ -26,7 +26,7 @@ class _CustomClass(object):
         st.booleans(),
         st.lists(st.text()),
         st.dictionaries(st.text(), st.integers()),
-        st.builds(_CustomClass, st.text()),
+        st.from_type(_CustomClass),
     ),
 )
 @example(None)
--- a/tests/test_primitives/test_container/test_base_container/test_pickle.py
+++ b/tests/test_primitives/test_container/test_base_container/test_pickle.py
@@ -26,7 +26,7 @@ class _CustomClass(object):
         st.booleans(),
         st.lists(st.text()),
         st.dictionaries(st.text(), st.integers()),
-        st.from_type(_CustomClass),
+        st.builds(_CustomClass, st.text()),
     ),
 )
 @example(None)
--- a/tests/test_primitives/test_container/test_base_container/test_pickle.py
+++ b/tests/test_primitives/test_container/test_base_container/test_pickle.py
@@ -1,0 +1,36 @@
+import pickle  # noqa: S403
+from typing import Any
+
+from hypothesis import example, given
+from hypothesis import strategies as st
+
+from returns.primitives.container import BaseContainer
+
+
+class _CustomClass(object):
+    def __init__(self, inner_value: Any) -> None:
+        self.inner_value = inner_value
+
+    def __eq__(self, other: Any) -> bool:
+        return (
+            type(other) == type(self) and  # noqa: WPS516
+            self.inner_value == other.inner_value
+        )
+
+
+@given(
+    st.one_of(
+        st.integers(),
+        st.floats(allow_nan=False),
+        st.text(),
+        st.booleans(),
+        st.lists(st.text()),
+        st.dictionaries(st.text(), st.integers()),
+        st.builds(_CustomClass, st.text()),
+    ),
+)
+@example(None)
+def test_pickle(container_value: Any):
+    """Ensures custom pickle protocol works as expected."""
+    container = BaseContainer(container_value)
+    assert pickle.loads(pickle.dumps(container)) == container  # noqa: S301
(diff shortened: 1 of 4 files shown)


tests/test_primitives/test_container/test_base_container/test_pickle.py
```diff
@@ -0,0 +1,35 @@
...
+    value: Any  # noqa: WPS110
+
+    def __init__(self, value: Any) -> None:  # noqa: WPS110
+        self.value = value  # noqa: WPS110
+
+    def __eq__(self, other: Any) -> bool:
+        return (
+            type(other) == type(self) and  # noqa: WPS516
+            self.value == other.value
+        )
```
sobolevn, line 19:
```suggestion
class _CustomClass(object):
    def __init__(self, inner_value: Any) -> None:
        self.inner_value = inner_value

    def __eq__(self, other: Any) -> bool:
        return (
            type(other) == type(self) and
            self.inner_value == other.inner_value
        )
```

tests/test_primitives/test_container/test_base_container/test_pickle.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@pytest.mark.parametrize('container_value', [
```
sobolevn, line 22: You can better use hypothesis here. A perfect case for it!

tests/test_primitives/test_container/test_base_container/test_pickle.py
```diff
@@ -0,0 +1,36 @@
...
+        st.integers(),
+        st.floats(allow_nan=False),
+        st.text(),
+        st.booleans(),
+        st.lists(st.text()),
+        st.dictionaries(st.text(), st.integers()),
+    ),
+)
+@example(None)
+@example(_CustomClass(42))
```
sobolevn, line 32: Why not `st.one_of(..., st.builds(_CustomClass))`? And `st.none()`? https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.none

thepabloaguilar: I saw `st.none()` but I wanted to guarantee `None` value will be always present since there's a lot of strategies in the `on_of` already and hypothesis doesn't guarantee it'll draw all the strategies!

sobolevn (approved):
Thank you! 

I will release a new version once I will be able to update mypy.
Right now it is broken due to `MYPYPATH` change.

tests/test_primitives/test_container/test_base_container/test_pickle.py
```diff
@@ -0,0 +1,36 @@
...
+
+@given(
+    st.one_of(
+        st.integers(),
+        st.floats(allow_nan=False),
+        st.text(),
+        st.booleans(),
+        st.lists(st.text()),
+        st.dictionaries(st.text(), st.integers()),
+        st.from_type(_CustomClass),
```
sobolevn, line 29:
```suggestion
        st.builds(_CustomClass, st.text()),
```

PR #1588, opened 2023-04-09
Makes `returns` compatible with `mypy` 1.2

thepabloaguilar:
# Makes `returns` compatible with `mypy` 1.2

During the upgrade some errors appeared which are kind ignored now but I'm going to open the respective issue which is:

- We can't have different Type Var variants (covariant, contravariant) than our super class, for now just a `type: ignore` is great 'cause we don't break any compatibility.


thepabloaguilar:
@sobolevn do you know why the error below is appearing when running the type safety tests?
```text
.../python3.10/site-packages/mypy/typeshed/stdlib/builtins.pyi:1104: error: Class cannot subclass "Any" (has type "Any")  [misc]
```

Note that we already have this in our `setup.cfg`:
```cfg
[mypy-typeshed.stdlib.unittest.mock]
disallow_subclassing_any = False
```
> Also, I've trued to change to `mypy-typeshed.stdlib.builtins` but haven't worked

sobolevn (commented):
Thanks A LOT!

There are two comments from me:
1. Tests are now run for 2 hours. I think we can disable recursive aliases
2. We can just disable `Any` subclass error completely for now

thepabloaguilar:
@sobolevn the error we're getting now it's an intermittent one, the test expects:

```text
main:19: note: Revealed type is "def [B, A] (*, b: builtins.list[B`-1], c: builtins.list[A`-2]) -> Union[A`-2, B`-1]"
```

But it's getting:
```text
main:19: note: Revealed type is "def [A, B] (*, b: builtins.list[B`-1], c: builtins.list[A`-2]) -> Union[A`-2, B`-1]"
```

Just the TypeVar order inside parenthesis are different, and as I've investigated it isn't something from our side the "error" is coming from the return of `expand_type` call in `CallableInference.from_usage(...)` method!

sobolevn:
Can you please ignore it for now and report to mypy? PR with the mypy fix is welcome.

sobolevn (commented):
Just one question :)

This look amazing! You did a great job, thank you!

sobolevn (approved):
🎉

sobolevn:
Feel free to merge when ready.

PR #1605, opened 2023-04-26
Add mypy-plugin extra dependency

zhukovgreen:
use `pip install classes[compatible-mypy]` for installing additional dependencies required by the mypy-plugin.

# I have made things!

## Checklist

## Related issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,11 @@ You are also required to [configure](https://returns.readthedocs.io/en/latest/pa
 `mypy` correctly and install our plugin
 to fix [this existing issue](https://github.com/python/mypy/issues/3157):
 
+```bash
+pip install returns[compatible-mypy]
+```
+and
+
 ```ini
 # In setup.cfg or mypy.ini:
 [mypy]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,6 @@ trio = "^0.22"
 attrs = "^23.1"
 httpx = "^0.24"
 
-mypy = "^1.4.0"
 wemake-python-styleguide = "^0.17"
 flake8-pytest-style = "^1.6"
 flake8-pyi = "^23.4"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ _ = "returns.contrib.hypothesis._entrypoint"
 python = "^3.7"
 
 typing-extensions = ">=4.0,<5.0"
+mypy = { version = "^1.4.0", optional = true }
 
 [tool.poetry.dev-dependencies]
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -48,6 +48,7 @@ _ = "returns.contrib.hypothesis._entrypoint"
 python = "^3.7"
 
 typing-extensions = ">=4.0,<5.0"
+mypy = { version = "^1.2.0", optional = true }
```
sobolevn, line 51:
```suggestion
mypy = { version = "^1.2", optional = true }
```

sobolevn (commented):
I would prefer this name: https://github.com/typeddjango/django-stubs#installation

zhukovgreen:
Should be fine now

README.md
```diff
@@ -40,6 +40,11 @@ You are also required to [configure](https://returns.readthedocs.io/en/latest/pa
 `mypy` correctly and install our plugin
 to fix [this existing issue](https://github.com/python/mypy/issues/3157):
 
+```bash
+pip install classes[compatible-mypy]
```
sobolevn, line 44:
```suggestion
pip install returns[compatible-mypy]
```

pyproject.toml
```diff
@@ -48,6 +48,7 @@ _ = "returns.contrib.hypothesis._entrypoint"
 python = "^3.7"
 
 typing-extensions = ">=4.0,<5.0"
+mypy = { version = "^1.2", optional = true }
```
sobolevn, line 51: I think we can now remove `mypy` from `dev-dependencies`

README.md
zhukovgreen: Resolved

pyproject.toml
zhukovgreen: Why do you think so? In this case, `poetry install` won't install mypy.

zhukovgreen: I was checking and if we remove the mypy from the dev-dependencies, then in order to let the python installed, we need to do:
```bash
$ poetry --only main --extras compatible-mypy
```

zhukovgreen: Some small example https://asciinema.org/a/N6NpAQsKGc6T4xR6DOfDQFxDn

zhukovgreen: https://asciinema.org/a/VOKGDmAFQBA8CgaZkleKvVQs8

sobolevn (commented):
Please, update mypy to `^1.4.0` and I will be happy to merge it!

zhukovgreen:
@sobolevn seems like all done now

PR #1610, opened 2023-05-02
returns with pyproject.toml

RANJITHROSAN17:


pyproject.toml
```diff
@@ -37,15 +37,13 @@ classifiers = [
...
-[tool.poetry.plugins.hypothesis]
-_ = "returns.contrib.hypothesis._entrypoint"
-
+[tool.poetry.plugins]
+pytest11 = {version = "^0.14", marker = "returns"}
+hypothesis = {version = "^4.51", marker = "returns"}
 
 [tool.poetry.dependencies]
 python = "^3.7"
+returns = "^0.20.0"
```
sobolevn, line 46: This does not seem right to me. How can the project be dependent on itself?

PR #1639, opened 2023-06-05
fixed HKT `to_str` example in documentation

### LINKED ISSUE #1638: mypy error for HKT example from docs
# Bug report

I was going through your documentation and stumbled upon an issue in HKT example: https://returns.readthedocs.io/en/latest/pages/hkt.html#id1.

## What's wrong

Trying out your `to_str` definition:

```python
from typing import TypeVar
from returns.maybe import Maybe
from returns.primitives.hkt import Kind1, kinded
from returns.interfaces.container import Container1, ContainerN

T = TypeVar("T", bound=Container1)

@kinded
def to_str(container: Kind1[T, int]) -> Kind1[T, str]:
    return container.map(str)

reveal_type(to_str(Maybe.from_value(1)))
```

Produces `mypy` error: `Value of type variable "T" of "to_str" cannot be "Maybe[Any]"  [type-var]`.

Playing around, I found, that replacing `Container1` with `ContainerN` fixes the issue (why?).

```python
T = TypeVar("T", bound=ContainerN)
```

## How is that should be

Examples from the documentation should work :)

I'm not sure whether this is a defect, maybe I'm doing something wrong, but before submitting this one I've rechecked that my code matches the docs closely.

## System information

- `python` version: 3.11.1
- `returns` version: 0.20.0
- `mypy` version: 1.3.0
  - I'm using config from here: https://github.com/wemake-services/wemake-python-styleguide/blob/master/styles/mypy.toml

vchernetskyi993:
# Fixed HKT `to_str` example in documentation

## Checklist

## Related issues

Closes #1638

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,8 +33,7 @@ jobs:
 
     - name: Install poetry
       run: |
-        curl -sSL \
-          "https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py" | python
+        curl -sSL "https://install.python-poetry.org" | python
 
         # Adding `poetry` to `$PATH`:
         echo "$HOME/.poetry/bin" >> $GITHUB_PATH
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ incremental in minor, bugfixes only are patches.
 See [0Ver](https://0ver.org/).
 
 
+## 0.20.1 WIP
+
+### Bugfixes
+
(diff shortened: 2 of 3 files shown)


sobolevn (approved):
Thanks! Much better now!

sobolevn:
Can you please also fix the ci? https://python-poetry.org/docs/#installing-with-the-official-installer

vchernetskyi993:
> Can you please also fix the ci? https://python-poetry.org/docs/#installing-with-the-official-installer

Changed `install-poetry` URL per docs.

PR #1710, opened 2023-10-19
Bump mypy to 1.6.1

sobolevn:
Closes https://github.com/dry-python/returns/pull/1709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,14 @@ See [0Ver](https://0ver.org/).
 
 ## 1.0.0 WIP
 
-### Misc
+### Features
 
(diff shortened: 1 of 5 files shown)


andersk:
Bisecting mypy shows that both failures were first triggered by python/mypy@14418bc3d2c38b9ea776da6029e9d9dc6650b7ea:

- python/mypy#15837

sobolevn:
Yeap, I have to rework how `detach_callable` works now.

sobolevn:
Sure, please, go ahead :)

thepabloaguilar:
I've tried but I have no idea 😄 

Why do you think `detach_callable` is the problem?

PR #1870, opened 2024-07-13
Issue 1543. sync impure_safe with safe signatures

RomanMIzulin:
# I have synced signatures of @safe and @impure_safe functions

## Checklist

- [x ] I have double checked that there are no unrelated changes in this pull request (old patches, accidental config files, etc)
- [x ] I have created at least one test case for the changes I have made
- [ ] I have updated the documentation for the changes I have made
- [x ] I have added my changes to the `CHANGELOG.md`

## Related issues

- Closes #1543

PR #1877, opened 2024-07-18
Better exception inference for `@safe` and `@impure_safe`

sobolevn:
I got this idea when reviewing your PR :)

CC @RomanMIzulin 
Refs https://github.com/dry-python/returns/pull/1870

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ See [0Ver](https://0ver.org/).
 
 ### Features
 
+- Improve inference of `ResultLike` objects when exception catching
+  decorator is applied with explicit exception types
 - Add picky exceptions to `impure_safe` decorator like `safe` has. Issue #1543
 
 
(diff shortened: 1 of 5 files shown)


RomanMIzulin:
Honestly, I do not get the idea of having ResultE when there is a Result[ValueType,ErrorType] with ErrorType with second arg. Just for typing a block of code when we are sure there is an error I guess. But if code is not deep just Result is okey?

RomanMIzulin:
Is the goal of this MR to proxy specific Exception rather than just Exception?
@sobolevn

sobolevn:
> Honestly, I do not get the idea of having ResultE when there is a Result[ValueType,ErrorType] with ErrorType with second arg

Back in the day we didn't have typevar defaults, so I had to add one more type alias for the common case. Right now we can deprecate `*E` aliases and use `TypeVar(..., default=Exception)` for `ResultLike`

RomanMIzulin:
and update doc like here https://returns.readthedocs.io/en/latest/pages/result.html#aliases
@sobolevn maybe issue for that?

sobolevn:
@RomanMIzulin yes, sure, go ahead!

PR #1896, opened 2024-07-30
Add annotation to Unwrappable type

RomanMIzulin:
# I have added silly annotation to Unwrappable type for getting rid of Pyright error `reportUnknownVariableType`

Note: it requires _new_ version of Pyright

## Checklist

- [x ] I have double checked that there are no unrelated changes in this pull request (old patches, accidental config files, etc)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ See [0Ver](https://0ver.org/).
 - Improve inference of `ResultLike` objects when exception catching
   decorator is applied with explicit exception types
 - Add picky exceptions to `impure_safe` decorator like `safe` has. Issue #1543
-- Unwrappable annotated for suppressing reportUnknownVariableType when importing for Pyright checks
 
 ### Misc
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ See [0Ver](https://0ver.org/).
 - Improve inference of `ResultLike` objects when exception catching
   decorator is applied with explicit exception types
 - Add picky exceptions to `impure_safe` decorator like `safe` has. Issue #1543
+- Unwrappable annotated for suppressing reportUnknownVariableType when importing for Pyright checks
```
sobolevn, line 17:
```suggestion
```

sobolevn (approved):
Thanks! You can try the `TODO` note:

> # TODO: add overloads for specific types, so it can narrow them with `TypeIs`

Basically, the idea is that types are narrowed like:

```type
x: Maybe[int]
if is_successful(x):
    reveal_type(x)  # Some[int]
```

I cannot promise that it is going to work, but we can try :)

PR #1975, opened 2024-12-02
Bind composite errors

edodo1337:
# I have made things!

## Checklist

## Related issues
This pull request fixes the [issue ](https://returns.readthedocs.io/en/latest/pages/result.html#how-to-create-unit-objects) of error types not composing in chains by using Union, improving clarity and type safety. However, `Any` can still appear in some cases. Setting `Success` and `Failure` to use `_Nothing` or `NoReturn` instead of `Any` helps with type inference but breaks many existing tests. I’m not an expert in type theory, but personally, I’d find it very useful to see fully composed error types without `Any`

### DIFF
--- a/returns/result.py
+++ b/returns/result.py
@@ -119,9 +119,9 @@ class Result(  # type: ignore[type-var]
         self,
         function: Callable[
             [_ValueType],
-            Kind2['Result', _NewValueType, _ErrorType],
+            Kind2['Result', _NewValueType, _ErrorType | _NewErrorType],
         ],
-    ) -> 'Result[_NewValueType, _ErrorType]':
+    ) -> 'Result[_NewValueType, _ErrorType | _NewErrorType]':
         """
         Composes successful container with a function that returns a container.
 
--- a/returns/context/requires_context_ioresult.py
+++ b/returns/context/requires_context_ioresult.py
@@ -249,11 +249,11 @@ class RequiresContextIOResult(  # type: ignore[type-var]
             Kind3[
                 RequiresContextIOResult,
                 _NewValueType,
-                _ErrorType,
+                _ErrorType | _NewErrorType,
                 _EnvType,
             ],
         ],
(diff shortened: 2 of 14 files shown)


returns/result.py
```diff
@@ -119,9 +119,9 @@ def bind(
         self,
         function: Callable[
             [_ValueType],
-            Kind2['Result', _NewValueType, _ErrorType],
+            Kind2['Result', _NewValueType, _ErrorType | _NewErrorType],
```
sobolevn, line 122: The problem with this approach is that it works with `Exception` subtypes, but it does not work if you have `Result[int, int]` and `Result[int, str]`

How would you compose these two objects? `Result[int, int | str]` might not make any sense.

sobolevn (commented):
I am not sure about this feature :)

As I told you in the chat. 
But, anyway - thanks a lot for working on it! 

Do you see any alternatives?

returns/result.py
edodo1337: > How would you compose these two objects? `Result[int, int | str]` might not make any sense.
Thank you for the review! I see your point, but isn’t `int | str` better than dealing with `Any`? At least with `int | str`, we can use pattern matching to handle each case explicitly.

sobolevn (changes requested):
Ok, I tested this out. Right now I kinda agree that this makes a lot of inference better.

But, in case we are adopting it - we need to make sure that all `ResultLike` behave the same way. Can you please address this?

edodo1337:
@sobolevn  Added similar behavior for `ResultLike`:, `IOResult`, `FutureResult`, `RequiresContextResult`, and `RequiresContextIOResult ` as requested. Let me know if I got everything right or missed something.

sobolevn:
The CI is failing, please take a look.

edodo1337:
There was a linting issue, it should be fine now.

sobolevn (commented):
Please, don't forget about interfaces.

PR #2060, opened 2025-03-09
Make the `hypothesis` plugin check laws from user-defined interfaces too

pradeep90:
# I have made things!

Thanks for the thoughtful and convenient plugin for running interface laws.

I found, however, that the plugin was not running the laws for an interface I defined myself. So, I decided to add that behavior.

Main changes:

+ We run laws from ancestor classes outside the `returns` module. This is the user-visible change.
+ In the `hypothesis` plugin, for each property test, we patch the interface with the container strategy:
   - Earlier, we were patching all ancestor classes, regardless of whether they had any laws. (We can see that in a [test file before the fix](https://github.com/dry-python/returns/blob/f4cc763d3809e478382ec5834e388860a0cdaf83/tests/test_contrib/test_hypothesis/test_interface_resolution.py))
   - Now, we are being more precise and patching only classes that have laws. But there should be no user-visible change. It's just cleaner.

## Checklist

## Related issues

### DIFF
--- a/returns/primitives/laws.py
+++ b/returns/primitives/laws.py
@@ -1,5 +1,5 @@
 from collections.abc import Callable, Sequence
-from typing import ClassVar, Generic, TypeVar, final
+from typing import ClassVar, Final, Generic, TypeVar, final
 
 from returns.primitives.types import Immutable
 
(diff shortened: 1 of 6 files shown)


returns/primitives/laws.py
```diff
@@ -12,6 +12,8 @@
 #: Special alias to define laws as functions even inside a class
 law_definition = staticmethod
 
+LAWS_ATTRIBUTE: str = '_laws'
```
sobolevn, line 15:
```suggestion
LAWS_ATTRIBUTE: Final = '_laws'
```

sobolevn (commented):
Thanks! Just one minor suggestion.

sobolevn (approved):
Thanks!

PR #2066, opened 2025-03-12
Allow users to override hardcoded `hypothesis` strategies

pradeep90:
## Override all the things!

The previous PR (#2061) allowed users to specify the container strategy. However, there's *still* a snag: the plugin hardcodes special strategies for `TypeVar` and `Callable`. These cannot be overridden by users, which is a problem when a law needs some generic value other than the container under test.

So, this PR allows users to override any of the strategies used by the plugin. It decouples the test-generation code from the hardcoded strategies by using a first-class value of `Settings`. The hardcoded strategies are in the default values of  `Settings` and can be overridden without trouble.

## Rationale

Why accept extra strategies within `check_all_laws`? Why not ask the user to register the strategies with `hypothesis` in the user's test file?

Our `hypothesis` plugin doesn't run tests. It creates tests and sticks them in the user's module.

So, even if I try:

```py
with type_resolver.strategies_for_types({...}): 
     check_all_laws(MyContainer)
```

that sets the context only when *defining* the tests, not when *running* them. When the tests are run, the types are no longer mapped to the strategies. We could ask the user to set the context when `pytest` runs the tests, but that would be error-prone and tedious for the user. So, we have to set the context inside the created test. Hence this PR.

### Why accept extra strategies?

Because the alternative is to write them to the `hypothesis` global variable, and global variables are evil.

Why can't the user register a strategy for `KindN` directly? It uses `TypeVarTuple`, which `hypothesis` doesn't understand. So, `hypothesis` thinks `KindN` is not a "subclass" of `KindN[int, str, bool]`, because the number of type arguments don't match. Sad.

As a workaround, we need to write those kinds of laws as `x: Callable[[], KindN[F, ...]]` instead of `x: KindN[F, ...]`. That way, at least our `Callable` override can catch these types.

## Implementation

+ `laws.py`:
  + Replace the impure context managers with a pure `Settings` value.
  + Split `Settings` into a default value (with the hardcoded strategies) and a user-provided value. 
     - We merge these two values. (`Monoid` would have been useful for this - hint hint :) )
+ The rest is docs and tests.
   + I've documented `Settings` inline. I believe it will be auto-included in the docs page.

## Test Plan

Run `test_custom_strategy_for_callable.py` after adding print statements to see the generated examples. (I don't know of other ways to see the examples in this case, since there is so much dynamically added context.)

```shell
pytest returns/ tests/ -vv --log-level INFO -m "hypothesis" -rP --pdb -k "test_custom_strategy_for_callable"
```

### Before 

The parameter `container: KindN[_F_Applicative, ...]` is always bound to a `_Wrapper` even though `_Wrapper` is not even an `Applicative`! (This is because `_Wrapper <: KindN`, sadly.)

```
some_identity <Wrapper: None> container <Wrapper: None>
some_identity <Wrapper: 0> container <Wrapper: 0>
some_identity <Wrapper: -27749> container <Wrapper: -116>
some_identity <Wrapper: False> container <Wrapper: False>
some_identity <Wrapper: False> container <Wrapper: True>
some_identity <Wrapper: b''> container <Wrapper: b''>
```

### After

Pass in a strategy for `Callable`. Now, we get `container` as `Maybe`s and `Result`s.

```
some_identity <Wrapper: 0> container <Nothing>
some_identity <Wrapper: 25634> container <Some: None>
some_identity <Wrapper: -16697> container <Success: b'v\xf0'>
some_identity <Wrapper: -16697> container <Success: None>
some_identity <Wrapper: -95> container <Some: 3>
some_identity <Wrapper: -95> container <Some: 0>
some_identity <Wrapper: 2> container <Success: 2>
some_identity <Wrapper: 2> container <Some: 2>
some_identity <Wrapper: -15533> container <Some: 1.5>
some_identity <Wrapper: -15533> container <Some: 1.5>
some_identity <Wrapper: 3> container <Success: 0.0>
some_identity <Wrapper: 3> container <Some: 0.0>
some_identity <Wrapper: 5739> container <Failure: 12655>
```

## Checklist


docs/pages/contrib/hypothesis_plugins.rst
```diff
@@ -140,8 +108,26 @@ like ``Future``, ``ReaderFutureResult``, etc
 that have complex ``__init__`` signatures.
 And we don't want to mess with them.
 
-You can also register a custom strategy to be used when running your
-container's laws:
+Warning::
+  Checking laws is not compatible with ``pytest-xdist``,
+  because we use a lot of global mutable state there.
+  Please, use ``returns_lawful`` marker
+  to exclude them from ``pytest-xdist`` execution plan.
```
sobolevn, line 115: We can also make a note about running them on a single node.

tests/test_contrib/test_hypothesis/test_laws/test_custom_strategy_for_callable.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+other_strategies: dict[type[object], StrategyFactory] = {
+    Callable: _callable_factory  # type: ignore[dict-item]
+}
+
+check_all_laws(
+    _Wrapper,
+    container_strategy=st.builds(_Wrapper, st.integers()),
+    other_strategies=other_strategies,
```
sobolevn, line 160: maybe `register_strategies`? `other` feels a bit off to me.

sobolevn (commented):
Thanks, the work you are doing is really impressive. Please, let me know when to make a new release.

docs/pages/contrib/hypothesis_plugins.rst
pradeep90: Ah, this was an existing paragraph that I moved up. I don't have anything to add here.

tests/test_contrib/test_hypothesis/test_laws/test_custom_strategy_for_callable.py
pradeep90: Agreed. How about `type_strategies`?

pradeep90:
> Thanks, the work you are doing is really impressive. Please, let me know when to make a new release.

Thanks! I appreciate your speedy responses too. Really helps keep the cycle time low.

I'll test-drive it for another week or so, just to make sure I don't run into any more issues that need fixing :)

docs/pages/contrib/hypothesis_plugins.rst
```diff
@@ -150,22 +136,64 @@ container's laws:
...
-The ``container_strategy`` will be used only when running the tests generated
-by the ``check_all_laws`` call above. It will have no effect on any other
-property tests that involve ``Number``. You cannot use this argument together
-with ``use_init``.
+You can also register strategies for other types:
+
+.. code:: python
+
+
+  from hypothesis import strategies as st
```
sobolevn, line 144:
```suggestion

  from hypothesis import strategies as st
```

sobolevn (approved):
Thank you!

pradeep90:
> Please, let me know when to make a new release.

@sobolevn I didn't run into any more problems :) We should be good for a new release. Thanks!

sobolevn:
Awesome! Scheduled for tomorrow.

sobolevn:
`0.25.0` is released 🎉 
Thanks a lot for your contributions!

PR #2094, opened 2025-04-02
Update CHANGELOG and fix mypy compatibility for curry.partial

AdrianAcala:
- Added a bugfix entry in CHANGELOG for `curry.partial` compatibility with mypy 1.6.1+.
- Modified `inference.py` to use a local variable for argument inference context.
- Added mypy configuration file and a test for `partial` functionality in `test_partial.py`.

# I have made things!

## Checklist

- [ ] I have updated the documentation for the changes I have made (since this is just fixing dependency issues, I didn't think a documentation update was necessary)

## Related issues


test_partial/test_partial.py
```diff
@@ -0,0 +1,10 @@
+from returns.curry import partial
```
sobolevn, line 1: We use `typesafety/` folder for typing tests.
You can find `test_curry` there.

returns/contrib/mypy/_typeops/inference.py
```diff
@@ -82,13 +82,15 @@ def _infer_constraints(
             self._fallback.arg_names,
             lambda index: checker.accept(exprs[index]),
         )
+
+        arg_context = checker.argument_infer_context()
```
sobolevn, line 86: Sorry, how does this change fix the mypy compat? What even is the problem that we are fixing here?

AdrianAcala: Based off of this commit https://github.com/python/mypy/commit/0291ec90d46655d47fcf220be7eab8b5f7c035e7,

the infer_constraints_for_callable function was modified to require an additional arg_names parameter. This change, combined with how mypy processes function calls internally, caused the direct use of checker.argument_infer_context() as a parameter to fail

When the context is saved to a local variable first, mypy correctly recognizes it during parameter binding.

AdrianAcala:
@sobolevn , ready for your review whenever you have time. Thanks so much for the work you do 😃

PR #2109, opened 2025-04-11
Add lock to ReAwaitable for concurrent awaits

proboscis:
# I have made things!

## Checklist

- [x] I have double checked that there are no unrelated changes in this pull request
- [x] I have created a test case for the changes I have made (test_reawaitable_concurrency.py)
- [x] I have added my changes to the \

## Related issues

This PR addresses the issue where multiple concurrent awaits on the same ReAwaitable instance could lead to unexpected behavior. By adding a lock around the critical section in the _awaitable method, we ensure that the coroutine is only awaited once even when multiple tasks await the same ReAwaitable concurrently.

The PR includes a test case that demonstrates concurrent awaiting works correctly with the lock in place.

🙏 Please, if you or your company finds \ valuable, help us sustain the project by sponsoring it transparently on https://github.com/sponsors/dry-python.

### DIFF
--- a/returns/primitives/reawaitable.py
+++ b/returns/primitives/reawaitable.py
@@ -2,6 +2,8 @@ from collections.abc import Awaitable, Callable, Generator
 from functools import wraps
 from typing import NewType, ParamSpec, TypeVar, cast, final
 
+import anyio
+
 _ValueType = TypeVar('_ValueType')
 _AwaitableT = TypeVar('_AwaitableT', bound=Awaitable)
 _Ps = ParamSpec('_Ps')
@@ -48,10 +50,11 @@ class ReAwaitable:
 
     """
 
-    __slots__ = ('_cache', '_coro')
+    __slots__ = ('_cache', '_coro', '_lock')
 
     def __init__(self, coro: Awaitable[_ValueType]) -> None:
         """We need just an awaitable to work with."""
+        self._lock = anyio.Lock()
         self._coro = coro
         self._cache: _ValueType | _Sentinel = _sentinel
 
@@ -101,9 +104,10 @@ class ReAwaitable:
 
     async def _awaitable(self) -> _ValueType:
         """Caches the once awaited value forever."""
-        if self._cache is _sentinel:
-            self._cache = await self._coro
-        return self._cache  # type: ignore
+        async with self._lock:
+            if self._cache is _sentinel:
+                self._cache = await self._coro
+            return self._cache  # type: ignore
 
 
 def reawaitable(
--- a/tests/test_primitives/test_reawaitable/test_reawaitable_concurrency.py
+++ b/tests/test_primitives/test_reawaitable/test_reawaitable_concurrency.py
@@ -1,0 +1,20 @@
+import anyio
+import pytest
+
+from returns.primitives.reawaitable import ReAwaitable
+
+
+async def sample_coro():
+    await anyio.sleep(0.1)
+    return "done"
+
+@pytest.mark.anyio
+async def test_concurrent_awaitable():
+    reawaitable = ReAwaitable(sample_coro())
+
+    async def await_reawaitable():
+        return await reawaitable
+    async with anyio.create_task_group() as tg:
+        task1 = tg.start_soon(await_reawaitable)
+        task2 = tg.start_soon(await_reawaitable)
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ incremental in minor, bugfixes only are patches.
 See [0Ver](https://0ver.org/).
 
 
+## 0.25.1
+
+### Bugfixes
+
+- Adds lock to `ReAwaitable` to safely handle multiple concurrent awaits on the same instance
+
+
 ## 0.25.0
 
 ### Features


tests/test_primitives/test_reawaitable/test_reawaitable_concurrency.py
```diff
@@ -0,0 +1,21 @@
+import anyio
+import pytest
+
+from returns.primitives.reawaitable import ReAwaitable
+
+
+async def sample_coro():
+    await anyio.sleep(0.1)
```
sobolevn, line 8:
```suggestion
    await anyio.sleep(1)
```

give less chance for random failures.

tests/test_primitives/test_reawaitable/test_reawaitable_concurrency.py
```diff
@@ -0,0 +1,21 @@
...
+    return 'done'
+
+
+@pytest.mark.anyio
+async def test_concurrent_awaitable():
+    reawaitable = ReAwaitable(sample_coro())
```
sobolevn: Please, add a link to the original bug on github.

tests/test_primitives/test_reawaitable/test_reawaitable_concurrency.py
```diff
@@ -0,0 +1,21 @@
+import anyio
+import pytest
+
+from returns.primitives.reawaitable import ReAwaitable
+
+
+async def sample_coro():
```
sobolevn, line 7:
```suggestion
async def sample_coro() -> str:
```

tests/test_primitives/test_reawaitable/test_reawaitable_concurrency.py
```diff
@@ -0,0 +1,21 @@
...
+    await anyio.sleep(0.1)
+    return 'done'
+
+
+@pytest.mark.anyio
+async def test_concurrent_awaitable():
```
sobolevn, line 13:
```suggestion
async def test_concurrent_awaitable() -> None:
```

returns/primitives/reawaitable.py
```diff
@@ -2,6 +2,8 @@
 from functools import wraps
 from typing import NewType, ParamSpec, TypeVar, cast, final
 
+import anyio
```
sobolevn, line 5: It is not in list of production deps: https://github.com/dry-python/returns/blob/bbfc3d35c04d48afed65e76ddb561909e94e5bdb/pyproject.toml#L51-L57

Can we use `if`s to find the correct lock type?

sobolevn (commented):
Thanks!

returns/primitives/reawaitable.py
proboscis: I see, but unfortunately i am not familiar with other libraries than pure `asyncio`.
Can we just use `asyncio.Lock`?

proboscis: Thank you for the fast review!

proboscis: ah maybe something like this?
```
try: import
```

sobolevn: yes, this `try` is fine :)

tests/test_primitives/test_reawaitable/test_reawaitable_concurrency.py
```diff
@@ -0,0 +1,71 @@
...
+from returns.primitives.reawaitable import ReAwaitable, reawaitable
+
+
+# Fix for issue with multiple awaits on the same ReAwaitable instance:
+# https://github.com/dry-python/returns/issues/2108
+async def sample_coro() -> str:
+    """Sample coroutine that simulates an async operation."""
+    await anyio.sleep(
+        1
+    )  # Increased from 0.1 to reduce chance of random failures
```
sobolevn, line 13:
```suggestion
    await anyio.sleep(1)  
```

returns/primitives/reawaitable.py
```diff
@@ -2,6 +2,16 @@
 from functools import wraps
 from typing import NewType, ParamSpec, TypeVar, cast, final
 
+# Try to use anyio.Lock, fall back to asyncio.Lock
+try:
+    import anyio  # noqa: WPS433
+except ImportError:
```
sobolevn, line 8: I think that we might need:

```suggestion
except ImportError:  # pragma: no cover
```

tests/test_primitives/test_reawaitable/test_reawaitable_concurrency.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@pytest.mark.anyio  # noqa: WPS210
+async def test_reawaitable_decorator() -> None:
+    """Test the reawaitable decorator with concurrent awaits."""
+
+    async def test_coro() -> str:  # noqa: WPS430
+        await anyio.sleep(
+            1
+        )  # Increased from 0.1 to reduce chance of random failures
```
sobolevn, line 39:
```suggestion
        await anyio.sleep(1)
```

tests/test_primitives/test_reawaitable/test_reawaitable_concurrency.py
```diff
@@ -0,0 +1,71 @@
...
+    async def test_func() -> int:  # noqa: WPS430
+        return 1
+
+    coro = test_func()
+    target = ReAwaitable(coro)
+
+    # Test the representation
+    assert repr(target) == repr(coro)
+    # Ensure the coroutine is properly awaited
+    await target
```
sobolevn, line 71:
```suggestion
    assert await target == 1
```

sobolevn (commented):
Thanks! Last comments

proboscis:
Thank you! I hope this works.

returns/primitives/reawaitable.py
```diff
@@ -2,6 +2,16 @@
...
 
+# Try to use anyio.Lock, fall back to asyncio.Lock
+try:
+    import anyio  # noqa: WPS433
+except ImportError:  # pragma: no cover
+    import asyncio  # noqa: WPS433
+
+    Lock = asyncio.Lock
+else:
+    Lock = anyio.Lock
```
sobolevn, line 13: You can type it as a simple interface that only supports `async with`, probably like `contextlib.AbstractAsyncContextManager`

sobolevn (commented):
Looks like we would have to use a custom protocol :(

returns/primitives/reawaitable.py
```diff
@@ -1,6 +1,26 @@
...
+# Try to use anyio.Lock, fall back to asyncio.Lock
+# Note: anyio is required for proper trio support
+try:
+    import anyio  # noqa: WPS433
+except ImportError:  # pragma: no cover
+    import asyncio  # noqa: WPS433
+
+    Lock: AsyncLock = asyncio.Lock
+else:
+    Lock: AsyncLock = anyio.Lock
```
sobolevn, line 23:
```suggestion
    Lock: type[AsyncLock] = asyncio.Lock
else:
    Lock = anyio.Lock
```

returns/primitives/reawaitable.py
```diff
@@ -1,6 +1,26 @@
 from collections.abc import Awaitable, Callable, Generator
 from functools import wraps
-from typing import NewType, ParamSpec, TypeVar, cast, final
+from typing import NewType, ParamSpec, Protocol, TypeVar, cast, final
+
+
+class AsyncLock(Protocol):
+    """A protocol for an asynchronous lock."""
```
sobolevn, line 7:
```suggestion
    """A protocol for an asynchronous lock."""
    
    def __init__(self) -> None: ...
```

returns/primitives/reawaitable.py
proboscis: ah, i missed this, thank you

proboscis:
At this point I am not sure why the CI is failing, and I am also not sure whether CI should be passing or not.
Could you help me understand how to make this pr valid?

My another concern is that it seems Lock object implementation must be changed depending on the async context (asyncio,anyio,trio). Since I have no understanding of anyio and trio, it is hard for me to provide valid lock implementation for all cases.
Since we do not know what async context will be used until the coroutine gets awaited, I implemented a hack to determine current context to select a lock. However, I understand this is just a hack and not really a solution.

Issue #2191, opened 2025-07-09
Incorrect result type for `returns.curry.partial` with keyword arguments for positional parameters


Nifacy:
# Bug report

## What's wrong

When using `returns.curry.partial`, if you partially apply a positional parameter using a keyword argument, the resulting type is incorrect. Result type of callable allows passing the remaining arguments as positional, but this leads to a runtime error.

### Example

```python
def foo(x: int, *, y: int) -> None:
    ...

bar = returns.curry.partial(foo, x=1)
reveal_type(bar)
# => Mypy: "def (y: builtins.int)"
# => Expected: "def (*, y: builtins.int)"

bar(2)
# No errors from mypy, but got runtime error
# TypeError: foo() got multiple values for argument 'x'
```

## How is that should be

In this scenario, the remaining parameters should be strictly keyword-only. The resulting type should reflect this restriction, and passing a positional argument should be flagged as an error at type checking time.

## System information

- `python` version: `3.13.4`
- `returns` version: `0.25.0`
- `mypy` version: `1.15.0`

PR #2446, opened 2026-07-09
Fix partial mypy plugin: make positionals keyword-only after a keyword-bound positional (#2191)

apoorvdarshan:
## Summary

`returns.curry.partial`'s mypy plugin produced an unsound signature when a
positional parameter was applied by keyword. For example:

```python
def foo(x: int, y: int) -> None: ...

bar = partial(foo, x=1)
reveal_type(bar)  # was: "def (y: int)"  -> should be: "def (*, y: int)"
bar(2)            # accepted by mypy, but raises at runtime:
                  # TypeError: foo() got multiple values for argument 'x'
```

Because `partial` is backed by `functools.partial`, bound arguments are
prepended and bound keywords are merged. Once an earlier positional parameter
is bound by keyword, the remaining positional parameters can no longer be
reached positionally at runtime, they must be passed by keyword. The plugin
was copying them unchanged (still positional), so `bar(2)` type-checked but
failed at runtime.

## Fix

In `Functions.diff` (`returns/contrib/mypy/_typeops/transform_callable.py`),
after determining which parameters were applied, we now detect the first
positional parameter that was applied by keyword (i.e. outside the leading
positional prefix that `functools.partial` fills). Every remaining positional
parameter after that "hole" is converted to keyword-only
(`ARG_POS`/`ARG_OPT` -> `ARG_NAMED`/`ARG_NAMED_OPT`). Parameters bound
positionally, and functions with no keyword hole, are unaffected.

The revealed type for the example above becomes `def (*, y: int)`, and
`bar(2)` is now correctly reported as `Too many positional arguments`.

## Tests / Verification

Added three regression cases to
`typesafety/test_curry/test_partial/test_partial.yml`:

- `partial_positional_arg_by_keyword_makes_rest_kwonly` - the issue's case
  now reveals `def (*, second: float) -> str`.
- `partial_positional_arg_by_keyword_rejects_positional_call` - a positional
  call on the partial is now a type error.
- `partial_middle_positional_arg_by_keyword_makes_rest_kwonly` - only the
  parameters after the keyword hole become keyword-only; earlier positionals
  are preserved.

All three fail on `master` and pass with this change. The full
`typesafety/test_curry` and `typesafety/test_pipeline` suites pass (120
cases), and `flake8` (wemake-python-styleguide) and `ruff`/`ruff format` are
clean on the changed files. Also manually verified at runtime that the new
keyword-only signature matches actual `functools.partial` behaviour.

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ See [0Ver](https://0ver.org/).
 
 - Add `mypy>=1.19,<1.22` support
 
+### Bugfixes
+
+- Fixes `partial` mypy plugin inferring the wrong signature when a positional
+  argument is applied by keyword: the remaining parameters are now correctly
+  marked as keyword-only, so passing them positionally is a type error instead
+  of a runtime `TypeError`
+
 
 ## 0.27.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,7 @@ incremental in minor, bugfixes only are patches.
 See [0Ver](https://0ver.org/).
 
 
-## 0.28.0
-
-### Features
-
-- Add `mypy>=1.19,<1.22` support
+## 0.28.1 WIP
 
 ### Bugfixes
 
@@ -18,6 +14,13 @@ See [0Ver](https://0ver.org/).
   argument is applied by keyword: the remaining parameters are now correctly
   marked as keyword-only, so passing them positionally is a type error instead
   of a runtime `TypeError`
+
+
+## 0.28.0
+
+### Features
+
+- Add `mypy>=1.19,<1.22` support
 
 
 ## 0.27.0
--- a/returns/contrib/mypy/_typeops/transform_callable.py
+++ b/returns/contrib/mypy/_typeops/transform_callable.py
@@ -1,4 +1,4 @@
-from typing import ClassVar, final
+from typing import ClassVar, Final, final
 
 from mypy.nodes import (
     ARG_NAMED,
@@ -24,7 +24,16 @@ from returns.contrib.mypy._structures.args import FuncArg
 
 #: Kinds of arguments that consume the leftover positional or keyword
 #: arguments (``*args`` and ``**kwargs``) and therefore cannot be applied.
-_VARIADIC_KINDS: frozenset[ArgKind] = frozenset((ARG_STAR, ARG_STAR2))
+_VARIADIC_KINDS: Final = frozenset((ARG_STAR, ARG_STAR2))
+
+#: Kinds of arguments that can be passed positionally.
+_POSITIONAL_KINDS: Final = frozenset((ARG_POS, ARG_OPT))
+
+#: Maps a positional argument kind onto its keyword-only counterpart.
+_KEYWORD_ONLY_KINDS: Final = {
+    ARG_POS: ARG_NAMED,
+    ARG_OPT: ARG_NAMED_OPT,
+}
 
 
 def proper_type(
@@ -131,18 +140,6 @@ class Functions:
     For example, one can need a diff of two callables.
     """
 
-    #: Kinds of arguments that can be passed positionally.
-    _positional_kinds: ClassVar[frozenset[ArgKind]] = frozenset((
-        ARG_POS,
-        ARG_OPT,
-    ))
-
-    #: Maps a positional argument kind onto its keyword-only counterpart.
-    _keyword_only_kinds: ClassVar[dict[ArgKind, ArgKind]] = {
-        ARG_POS: ARG_NAMED,
-        ARG_OPT: ARG_NAMED_OPT,
-    }
-
     def __init__(
         self,
         original: CallableType,
@@ -192,7 +189,7 @@ class Functions:
         )
         seen_positional = 0
         for index, arg in enumerate(original_args):
-            if arg.kind not in self._positional_kinds:
+            if arg.kind not in _POSITIONAL_KINDS:
                 continue
             applied = self._was_applied(arg, index, intermediate_names)
             if applied and seen_positional >= positional_prefix:
@@ -202,7 +199,7 @@ class Functions:
 
     def _remaining_arg(self, arg: FuncArg, *, keyword_only: bool) -> FuncArg:
         """Copies an unapplied argument, making it keyword-only if needed."""
-        keyword_kind = self._keyword_only_kinds.get(arg.kind)
+        keyword_kind = _KEYWORD_ONLY_KINDS.get(arg.kind)
         if keyword_only and keyword_kind is not None:
             return FuncArg(arg.name, arg.type, keyword_kind)
         return arg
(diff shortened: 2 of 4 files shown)


returns/contrib/mypy/_typeops/transform_callable.py
```diff
@@ -128,36 +136,92 @@ def __init__(
         self._original = original
         self._intermediate = intermediate
 
-    def diff(self) -> CallableType:
+    #: Kinds of arguments that can be passed positionally.
+    _positional_kinds: ClassVar[frozenset[ArgKind]] = frozenset((
```
sobolevn: Please, move this before the `__init__`

typesafety/test_curry/test_partial/test_partial.yml
```diff
@@ -42,6 +42,52 @@
     reveal_type(partial(two_args, second=1.0))  # N: Revealed type is "def (first: int) -> str"
 
 
+- case: partial_positional_arg_by_keyword_makes_rest_kwonly
+  disable_cache: false
+  main: |
+    from returns.curry import partial
+
+    def two_args(first: int, second: float) -> str:
```
sobolevn, line 63: please, test the same with `*, second: float` :)

CHANGELOG.md
```diff
@@ -12,6 +12,13 @@ See [0Ver](https://0ver.org/).
 
 - Add `mypy>=1.19,<1.22` support
 
+### Bugfixes
```
sobolevn, line 11: This should go under new `## WIP` section, 0.28.0 was already released.

typesafety/test_curry/test_partial/test_partial.yml
```diff
@@ -42,6 +42,52 @@
...
+    # reported as keyword-only:
+    reveal_type(partial(two_args, first=1))  # N: Revealed type is "def (*, second: float) -> str"
+
+
+- case: partial_positional_arg_by_keyword_rejects_positional_call
+  disable_cache: false
+  main: |
+    from returns.curry import partial
+
+    def two_args(first: int, second: float) -> str:
```
sobolevn, line 90: Please, also add a test case with `first: int, /, second: int, third: str`, where you bind `second` as kw arg :)

returns/contrib/mypy/_typeops/transform_callable.py
```diff
@@ -128,36 +136,92 @@ def __init__(
...
+        return arg
+
+    def _was_applied(
+        self,
+        arg: FuncArg,
+        index: int,
+        intermediate_names: list[str | None],
+    ) -> bool:
+        """Tells whether an original argument was already applied."""
+        if arg.kind in {ARG_STAR, ARG_STAR2}:
```
sobolevn: we can move this to be a constant as well :)

apoorvdarshan:
Thanks for the review @sobolevn! I've addressed all five points in `a31ef11`:

1. **`transform_callable.py` — move `_positional_kinds` before `__init__`**: In `Functions`, both class vars (`_positional_kinds` and `_keyword_only_kinds`) now come first, and `__init__` follows them.

2. **`test_partial.yml` — test the same with `*, second: float`**: Added `partial_keyword_only_arg_stays_keyword_only`, which binds `first` on `def two_args(first: int, *, second: float)` and asserts the result stays `def (*, second: float) -> str`.

3. **`CHANGELOG.md` — should be under a new WIP section**: Moved the bugfix entry out of the released `0.28.0` and into a new `## 0.28.1 WIP` section at the top (matching the project's `## <version> WIP` convention). `0.28.0` now keeps only its released `mypy>=1.19,<1.22` feature.

4. **`test_partial.yml` — add `first: int, /, second: int, third: str` binding `second` as kw arg**: Added `partial_positional_only_before_keyword_bound_arg`. Binding `second` by keyword makes `third` keyword-only while the positional-only `first` (before the hole) stays positional-only, giving `def (int, *, third: str) -> str` (mypy renders the positional-only param without a name).

5. **`transform_callable.py` — move this to a constant as well**: Extracted `{ARG_STAR, ARG_STAR2}` into a module-level `_VARIADIC_KINDS` constant, now reused in both `_keyword_hole_at` and `_was_applied`.

Verified locally: the full `typesafety/test_curry/test_partial/` suite passes (16/16, incl. the two new cases) on mypy 1.20, `mypy returns/contrib/mypy/` is clean, and `ruff` + `flake8` pass on the changed file. The original #2191 fix behavior is unchanged.

returns/contrib/mypy/_typeops/transform_callable.py
```diff
@@ -14,6 +22,10 @@
 
 from returns.contrib.mypy._structures.args import FuncArg
 
+#: Kinds of arguments that consume the leftover positional or keyword
+#: arguments (``*args`` and ``**kwargs``) and therefore cannot be applied.
+_VARIADIC_KINDS: frozenset[ArgKind] = frozenset((ARG_STAR, ARG_STAR2))
```
sobolevn:
```suggestion
_VARIADIC_KINDS: Final = frozenset((ARG_STAR, ARG_STAR2))
```

returns/contrib/mypy/_typeops/transform_callable.py
```diff
@@ -119,6 +131,18 @@ class Functions:
     For example, one can need a diff of two callables.
     """
 
+    #: Kinds of arguments that can be passed positionally.
+    _positional_kinds: ClassVar[frozenset[ArgKind]] = frozenset((
```
sobolevn: So, maybe this should be a constant as well? It is strange that something a module level constant and something is a class level attr.

typesafety/test_curry/test_partial/test_partial.yml
```diff
@@ -42,6 +42,79 @@
...
+  disable_cache: false
+  main: |
+    from returns.curry import partial
+
+    def two_args(first: int, *, second: float) -> str:
+        ...
+
+    # `second` is already keyword-only, so binding the positional `first`
+    # by keyword leaves it keyword-only just as before:
+    reveal_type(partial(two_args, first=1))  # N: Revealed type is "def (*, second: float) -> str"
```
sobolevn, line 82: Please, also add `partial(two_args, second=2)` case

apoorvdarshan:
Thanks for the follow-up review! Addressed all three in `72787b17`:

1. `_VARIADIC_KINDS` now uses your suggested form: `_VARIADIC_KINDS: Final = frozenset((ARG_STAR, ARG_STAR2))` (imported `Final` from `typing`).
2. Agreed it was inconsistent to mix module-level constants and class-level attrs. Promoted the two attrs I'd added on `Functions` to module-level `Final` constants next to `_VARIADIC_KINDS` — `_POSITIONAL_KINDS` (frozenset) and `_KEYWORD_ONLY_KINDS` (the kind→kw-only mapping) — and updated their references. Behavior is unchanged.
3. Added a `partial(two_args, second=2)` case (`partial_last_arg_by_keyword_keeps_first_positional`): binding the last parameter by keyword leaves no positional after the hole, so `first` stays positional → `def (first: int) -> str`.

Verified locally with the pinned mypy: `mypy returns/contrib/mypy/` is clean, the full `typesafety/test_curry/test_partial/` suite passes (incl. the new case), and ruff is clean on the changed file.

The earlier `Run benchmarks (memory)` red was a transient CodSpeed installer 504 (infra, not the change); this push should re-run CI.

sobolevn (approved):
Thank you!

